PR #38, opened 2018-03-14
Draft of messages format

pcppcp:


matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
+Raiden transport messages
```
LefterisJP: Shouldn't these go to the already existing `transport.rst` ?

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+#########################
+
+Definitions
+===========
+* Sender: entity sending the message
+* Recipient: entity able of receiving and decoding the message
```
LefterisJP:
```diff
- able of receiving and decoding 
+ able to receive and decode
```

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
+Raiden transport messages
+#########################
+
+Definitions
```
LefterisJP: These should not be here. We have a terminology document. Please simply use terms from there or add new ones there if they don't already exist.

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+Envelope fields:
+===============
+* signature - a sender’s signature of the data field
+* content-type - how the data are encoded, i.e.:
+    * text-plain: a simple escaped JSON stored as a JSON string
+    * base-64: base-64 encoded JSON
+* data - tranmitted data
+Message MUST contain all mentioned fields and MUST be discarded by the Recipient otherwise.
+
+After unpacking `data` using the method(s) specified in content-type field, recipient MUST obtain a JSON object of a following format:
```
LefterisJP: `a following` -> `the following`

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+===========
+* Sender: entity sending the message
+* Recipient: entity able of receiving and decoding the message
+* Message: A data object exchanged between Sender and Recipient
+* Message Field: an attribute of JSON data format that is required to be included in the message
+
+Transport envelope
+==================
+Messages are exchanged in JSON format.
+Message itself contains metadata that allow recipient to verify sender of the message (via use of PK signatures).
```
palango: A message

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+            'locksroot': '0x0000000000000000000000000000000000000000000000000000000000000000',
+            'extra_hash': '0x647d722644d0dc43da475586ec9150e1d26ac0aceedb3b642d50c0b023d8774e',
+            'signature': '0x6ef31b0aa19e093721419d8e0eabe11e67824a8daa3549c99e7fa0992d43de022181cefe0fe0c91855bca40f1a73287ec95e690d59e64ce9ebd436cf834bbd7f1b'
+        }
+    }
+}
+``
+
+
+See also:
```
LefterisJP: Make this into a proper RST link.

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+{
+    'signature': '0xabcd....01234',
+    'content-type’: 'base-64',
+    'data': 'eyAnaGVhZGVyJzogeyAndHlwZScgOiAndGV4dC1tZXNzYWdlJywgJ3NlbmRlcic6ICAnMHhhYWVlLi4uMTIzNCcsICd0aW1lc3RhbXAnOiAxMjM0NTY3ODkgfSwNCidib2R5JzogeyAnbWVzc2FnZSc6ICdoZWxsbyB3b3JsZCcgfX0='
+}
+``
+
+Envelope fields:
+===============
+* signature - a sender’s signature of the data field
```
palango: senders (without ')

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+* content-type - how the data are encoded, i.e.:
+    * text-plain: a simple escaped JSON stored as a JSON string
+    * base-64: base-64 encoded JSON
+* data - tranmitted data
```
palango: transmitted

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+    'signature': '0xabcd....01234',
+    'content-type’: 'base-64',
+    'data': 'eyAnaGVhZGVyJzogeyAndHlwZScgOiAndGV4dC1tZXNzYWdlJywgJ3NlbmRlcic6ICAnMHhhYWVlLi4uMTIzNCcsICd0aW1lc3RhbXAnOiAxMjM0NTY3ODkgfSwNCidib2R5JzogeyAnbWVzc2FnZSc6ICdoZWxsbyB3b3JsZCcgfX0='
+}
+``
+
+Envelope fields:
+===============
+* signature - a sender’s signature of the data field
+* content-type - how the data are encoded, i.e.:
```
palango: What is the idea behind allowing different content types? IOW why not define one content type?

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+* JSON format is widely used and can be easily validated using JSON-schema (http://json-schema.org/)
+* JSON is easily extensible and extension can be made backward-compatible
+* Message format must not be dependent on the transport used - it should not matter whether the message is sent via Matrix or REST
+* Sender identity verification must not be part of the transport. It should be possible to submit the message over an untrusted channel or delegate.
+
+Example of a message
+-------------------
+``
+{
+    'signature': '0xabcd....01234',
```
palango: Does a signature allow for check-summing of the message? Do you think that a checksum would be useful to have?

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+    }
+    'body': { <JSON OBJECT> }
+}
+``
+
+Header fields
+=============
+* Sender (string) - public key used to verify signature of the message. Recipient MUST validate the signature and MUST discard the message if the signature doesn’t match
+* Timestamp (int) - time the message was sent (note: a block hash might also be used to enforce timestamp tampering)
+* Type (string) - user-specified type of the message
```
palango: Is there a list of the possible types?

matrix_messages.rst
pcppcp: When debugging it's easier to use plaintext, but to save bandwidth (i.e. with IoT) it might be beneficial to have an option for compressed content-type.

matrix_messages.rst
pcppcp: we don't need that, the signature is also a checksum (data to sign are hashed first, then signed; vice versa when verifying the signature)

matrix_messages.rst
pcppcp: not yet, but we can reuse what's in Raiden already:

```
    'Ack',
    'Ping',
    'SecretRequest',
    'Secret',
    'DirectTransfer',
    'Lock',
    'LockedTransfer',
    'MediatedTransfer',
    'RefundTransfer',
```

matrix_messages.rst
pcppcp: yea, I can move it there if we agree on message format. On the other hand transport document should maybe only contain low-level transport specification and messages SPEC may be a separate document.

matrix_messages.rst
pcppcp: Noted; I'll move it there once we settle on format.

matrix_messages.rst
```diff
@@ -0,0 +1,92 @@
...
+* content-type - how the data are encoded, i.e.:
+    * text-plain: a simple escaped JSON stored as a JSON string
+    * base-64: base-64 encoded JSON
+* data - transmitted data
+Message MUST contain all mentioned fields and MUST be discarded by the Recipient otherwise.
+
+After unpacking `data` using the method(s) specified in content-type field, recipient MUST obtain a JSON object of the following format:
+``
+{
+    'header': {
```
andrevmatos: I'd like to discuss if a full/nested envelope like this is really needed. While agreeing and not getting away from the rationale, I'd rather push for a simpler approach, bundling as few informations as possible, something that is as agnostic as possible, while keeping the security model. Redundant information is not useful.
1. sender shouldn't be in the message. It should be infered from other means (matrix `userId`, but should be easily replaced by other uniquely identifying sources, transport agnostic like on-chain data [e.g channel open event], or even challenge identity on p2p protocols). It'll be validated anyway through `ecrecover` or equivalent, so can't see the point of enforcing it in the message layer.
2. timestamp shouldn't be bundled. It should be infered through message receiving event (both client-side or server provided). For the cases  where it should be sender validated/signed (can't think on any right now), it should be part of the msg/payload body.
3. Type also should be inside msg/payload where required, as above.
4. Can't see why a content-type/encoding is required. Better if it's just plain-text string (easier debugging), with signature validated before decoding. Where privacy is important, public-key exchange must be performed, and encrypted message sent as payload instead.

My proposal is: we can also simply enforce that signed data (payload) is appended with some kind of unique identification of the transport endpoint. e.g:
- for matrix, if `msg = json.dumps(data)` is the actual payload, `sig = sign(f"{msg}\n{userId}")`, `room.sendText(f"{msg}\n{sig}")`, so receiver can do `ecrecover(f"{msg}\n{sender.userId}", sig)` (receiver append userId by itself, before checking signature). It may be an additional layer of replay-protection, but the actual layer should be in the payload data itself.
-  for any other p2p protocol, a unique id should always be available. Even in the lowest level (e.g. UDP), the connect endpoint could be used for it, so it's not tied to matrix protocol in any way.

My point is just to avoid bundling redundant information in the message. It should contain mostly useful data only. Rationale is the metaphorically same for database primary-key generation: if you can use a unique column, prefer it over adding a meaningless ID.

andrevmatos: [PoC implementation of my proposal above](https://github.com/andrevmatos/matrix-test/blob/97cc48e5d45c6d7e6bce9695bb8fa37fa190b673/main.py#L109-L165)

pcppcp: - user id should be nothing else, but ethereum address of the Raiden node that signed the message. This id is easily verifiable on-chain. (if for example a balance proof is submitted the server may choose to ignore it if the signer is not a Participant of the channel in question)
- using another layer for identity also makes the whole thing harder to implement and reason about (think about possible security problems if the identity service gets compromised)
- the messaging must thus work even without id (use case: submitting balance proof with REST)
- From Raiden Node perspective transport layer must be trustless. Its security, if any, is just an added benefit, but it should not replace security of the most important layer - ethereum blockchain. Compromise of the auxiliary services (like Matrix servers or Raiden app) must in worst case cause a DoS or loss of Raiden node operator funds only. 
- there's rarely any advantage in inventing a new message format, especially unstructured one. I chose JSON because of its obvious benefits - it's widely used, has validation schema standard that can be used to do basic semantic check of both the envelope and data contents and is extensible enough while maintaining backward compatibility.
- I actually prefer block hash to timestamp as it can't be faked to make the message come from the future. There might be possible timing attacks and the service should be very conservative which messages will it actually act on. It also makes discarding the messages easy - the server might choose to ignore very old messages (possibly sent by a malicious party to DoS the service).
- type _can_ also be included in data format, it's just easier to validate the message if it's contained in the header.
- for content-type see answer above, it might be beneficial to reduce bandwidth in certain cases (though this will be probably handled by underlying protocol by compressing the payload)

> for any other p2p protocol, a unique id should always be available

This makes the messaging protocol-depenent, that's not what we want

> Where privacy is important, public-key exchange must be performed, and encrypted message sent as payload instead.

Encryption should be done by protocol/network layer, let's not invent the wheel here.

> It'll [sender address] be validated anyway through ecrecover or equivalent, so I can't see the point of enforcing it in the message layer.

I somewhat agree (if signatures will be always ec-recoverable). From debugging perspective it's easier to have an option of including additional data in the message. The field can be ignored by the reciver though.

> My point is just to avoid bundling redundant information in the message. It should contain mostly useful data only

there's no added benefit in having the format restricted, rather than extensible. The receiver may use only a small subset of the fields, but it shouldn't be a problem if the message contains more data (Robustness principle)

andrevmatos: Well, almost everything you said can be achieved mostly through my simplified proposal. Notice my proposal also uses JSON fully, I don't want to invent some special format, and all validation is mandatorily coupled with ETH address.
My point in adding some userId to the signed message is not to tie it to the protocol, but rather just add some information to the signed data so the signature include the sender _commitment_ with that userId. It's not central to my proposal, and can be discarded.
The point is, if we want to be fully protocol agnostic, it can be simply `msg = json.dumps(data); sig = sign(msg);  room.send_text(msg + '\n' + sig);`, and that's it. I don't want to restrict the format of the messages, actually the other way around, all data should be JSON-encoded and signed (as above), and receiver can validate (ecrecover) it, but choose to care only about the fields it wants. Plain-text JSON is better compressed than e.g. base64.

matrix_messages.rst
hackaugusto: Tthis is invalid json btw, json uses double quotes and not single quotes

LefterisJP:
@andrevmatos if you want someone to review a PR please mention or assign PR to them instead of mentioning at the review part because damn github doesn't generate a notification otherwise.

andrevmatos:
@LefterisJP I think you guys already were reviewers of the issue, and it's @pcppcp's issue, so I didn't ask for review. But, please, look at the comment above, and say what you think.

LefterisJP:
@andrevmatos I will. But I was referring to you mentioning us at the "review comment". The one you make when you approve/reject a PR. Those mentions don't seem to send out notifications.

ulope:
During our call last week we settled on using @andrevmatos's solution (`<message>\n<signature>`) for the initial PoC implementation in Raiden since that allows us to get a working round-trip capable sytem relatively fast by re-using the existing binary encoded messages. 

In parallel we will investigate if using canonical json with a minimal envelope and content spec will fit our needs for the other services and the "proper" raiden implementation.

PR #42, opened 2018-03-26
updating some details, adding high level description of pathfinding

err508:


pathfinding_service.rst
```diff
@@ -18,6 +18,16 @@ Assumptions
 * Every pathfinding service is responsible for a single token network. Pathfinding services are scaled on process level to handle multiple token networks.
 
 
+High-Level-Description
+======================
+A node can request a list of possible paths from start point to endpoint for a given transfer value.
+The get_paths method implements canonical Dijkstra to return a given number of paths for a mediated transfers of a given value. The design regards the Raiden network as unidirectional weighted graph, where the default weights (and therefore the primary constraint of the optimization) are the fee's of each channel. Additionally we applied two heuristics to quantify two desirable properties of the resulting graphs:
```
palango: Dijkstra algorithm

palango: fee's -> fees

palango: I think we should remove one of the two *two*s in the last sentence.

pathfinding_service.rst
```diff
@@ -18,6 +18,16 @@ Assumptions
...
+
+i) a hard coded parameter DIVERSITY_PEN_DEFAULT defined in the config; this value is added to each edge that is part of a returned path as a bias. This results in an output of "pseudo-disjoint" paths, i.e. the optimization will prefer paths with a minimal edge intersection. This should enable nodes to have a suitable amount of options for their payment routing in the case some paths are slow or broken. However, if a node has only one channel (i.e. a light client) payments could be routed through, the method will still return the specified number of paths.
```
palango: Start with big A

pathfinding_service.rst
```diff
@@ -161,13 +177,16 @@ Example
         "error": "Invalid signature."
     }
 
-``api/1/paths``
-^^^^^^^^^^^^^^^
+``api/1/get_paths``
```
palango: why do we rename these?

pathfinding_service.rst
```diff
@@ -161,13 +177,16 @@ Example
...
+^^^^^^^^^^^^^^^^^^^
 
-Request a list of possible paths from startpoint to endpoint for a given transfer value.
+The method will do num_path iterations of Dijkstra on the last-known state of the Raiden Network (regarded as directed weighted graph) to return num_path different path's for a mediated transfer of _value_.
```
palango: I'd prefer Dijkstra's algorithm

palango: pls mark num_paths  as code: `num_paths`

pathfinding_service.rst
```diff
@@ -306,4 +325,4 @@ Example
     {
         "error": "No suitable path found."
     }
-
+         0
```
palango: ?

pathfinding_service.rst
```diff
@@ -18,6 +18,16 @@ Assumptions
 * Every pathfinding service is responsible for a single token network. Pathfinding services are scaled on process level to handle multiple token networks.
 
 
+High-Level-Description
+======================
+A node can request a list of possible paths from start point to endpoint for a given transfer value.
+The get_paths method implements canonical Dijkstra algorithm to return a given number of paths for a mediated transfers of a given value. The design regards the Raiden network as unidirectional weighted graph, where the default weights (and therefore the primary constraint of the optimization) are the fees of each channel. Additionally we applied two heuristics to quantify desirable properties of the resulting graphs:
```
palango: - implements *the*/*a* canonical Dijkstra
- for a mediated transfer (remove the s)
- as *an* unidirectional
- properties of the resulting graph (should be just one, right?)

pathfinding_service.rst
```diff
@@ -18,6 +18,16 @@ Assumptions
...
+
+
+ii) The second heuristic is configurable via the optional argument 'bias', which models the trade-off between speed and cost of mediated transfer; with default 0, get_paths will  optimize with respect to overall fees only (i.e. the cheapest path). On the other hand, with bias=1, get_paths will look for path's with the minimal number of hops (i.e. the  -theoretical - fastest path). Any value in [0,1] is accepted, an appropriate value depends on the average channel_fee in the network (in simulations mean_fee gave decent results for the trade-off between speed and cost). The reasoning behind this heuristic is that a node may have different needs, w.r.t to good to be paid for - buying a potato should be fast, buying a yacht should incorporate low fees.
```
palango: - will look for *paths* with the

pathfinding_service.rst
```diff
@@ -161,13 +177,16 @@ Example
         "error": "Invalid signature."
     }
 
-``api/1/paths``
-^^^^^^^^^^^^^^^
+``api/1/get_paths``
+^^^^^^^^^^^^^^^^^^^
 
-Request a list of possible paths from startpoint to endpoint for a given transfer value.
+The method will do ``num_path`` iterations of Dijkstra on the last-known state of the Raiden Network (regarded as directed weighted graph) to return num_path different path's for a mediated transfer of _value_.
```
palango: different *paths*

pathfinding_service.rst
```diff
@@ -161,13 +177,16 @@ Example
...
 
-This method will be rate-limited in a configurable way. If the rate limit is exceeded, clients can be required to pay the path-finding service with RDN tokens via the Raiden Network. The required path for this payment will be provided by the service for free. This enables a simple user experience for light users without the need for additional on-chain transactions for channel creations or payments, while at the same time monetizing extensive use of the API.
```
palango: What happened to the comment about rate-limiting and payments?

pathfinding_service.rst
err508: to make the naming consistent with the methods we implemented

pathfinding_service.rst
err508: We don't have implemented anything like this yet, but I can put it back in the spec

palango: Our idea was to add payment functionality after the initial prototype is built.

pathfinding_service.rst
palango: not sure about naming conventions, but I thought the get comes from `GET` verb in HTTP.

devium: Note: the `TokenNetwork` methods in the code are NOT the API interface. `TokenNetwork` is just the internal business logic.
The API interface that receives requests and converts them for the business logic is defined in `ServiceApi` and its registered `PathfinderResource` derivatives.

So that endpoint still is `'api/1/paths'`, called with a `GET`.

pathfinding_service.rst
devium: @err508 It's not implemented yet but we definitely need this rate limiting as pay-per-request imposes an unnecessary restriction for casual users. Paying with RDN would require a RDN balance and a RDN connection (be it µRaiden oder Raiden) to the pathfinding service. We don't want that for "casual" users, only for premium/high throughput users.

pathfinding_service.rst
err508: What I described in the spec is exactly that internal logic, the api call is not interesting to spec out I guess. I'd prefer consistency with the internal method, which tells you where to look, if you want to find in the code for what is described in the spec.

palango: In that [part](https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#public-endpoints) we explicitly define the REST API endpoints that will be used. So is shuld be `api/1/paths` here.

pathfinding_service.rst
err508: Added this as a section `future work`.

pathfinding_service.rst
err508: Okay done. Please see if my additions to spec make sense to you then or if they should go somewhere else.

PR #45, opened 2018-03-29
Add token network address to all api endpoints

palango:
The PFS should be able to handle multiple token networks.

In order to know which token network should be updated or for which token network paths are requested the PFS needs the know the target token network address.

Not sure about the payment stuff: It would make sense to pay more for a bigger token network, but that may complicate the payment process.

### DIFF
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -55,8 +55,8 @@ A path finding service must provide the following endpoints. The interface has t
 
 The examples provided for each of the endpoints is for communication with a REST endpoint.
 
-``api/1/balance``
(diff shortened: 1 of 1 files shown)


pathfinding_service.rst
```diff
@@ -108,7 +110,7 @@ Example
                 "secrethash": "<keccak-hash>"
             },
         ],
-    }'  /api/1/balance
+    }'  /api/0xtoken_network/1/balance
```
devium: Tiny typo here: `/api/1/0xtoken_network/balance`

palango: Fixed

PR #52, opened 2018-04-06
Draft of messages format

pcppcp:
New document describes basics of the message format discussed & proposed on Thursday's call.

Issues to discuss
- should the document contain raiden-specific message definitions? Existing messages and data types are already described in `messaging.rst` and since the JSON is basically 1:1 copy of binary format, I guess this document can just mention that.
- the docs is rather short, but I don't honestly know what else I can add :)

### DIFF
--- a/json_encoding.rst
+++ b/json_encoding.rst
@@ -1,0 +1,55 @@
+#########################
+Raiden transport messages
+#########################
+
+Transport
+=========
+Messages are exchanged in JSON format using Matrix as a transport layer. Messages can be sent either via broadcast channel or via direct messaging
+channel.
+
+Rationale
+---------
+* plaintext JSON messages are easy to read for both humans and machines
+* JSON format is widely used and can be easily validated using JSON-schema (http://json-schema.org/)
+* JSON is easily extensible and extension can be made backward-compatible
+* Message format must not be dependent on the transport used - it should not matter whether the message is sent via Matrix or REST
+* Sender identity verification must not be part of the transport. It should be possible to submit the message over an untrusted channel or delegate.
+
+Fields validation
+-----------------
+* Receiver SHOULD use JSON schema to validate integrity of received message depending on the message type.
+* For messages that contain data later used as a smart contract parameters, receiver MUST check for overflows and conversion errors.
+* Any values to be used on-chain MUST NOT in any circuimstances be floats.
+* For ecrecover-able signatures, Sender identity SHOULD be verified.
+
+Simplest valid message is defined as follows
+
+``
+{}
+``
+
+
+Example of a message
+-------------------
+
+``
+{
+    'type': 'SubmitBalanceProof',
+    'network_id': 1,
+    'balance_proof': {
+        'signature': '0x123...abc',
+        'nonce': 1
+    },
+    'reward_proof': {
+        'signature': '0x234...def',
+        'reward_amount': 100,
+        'channel_id': 23
+    }
+    }
+}
+``
+
+See also `Messaging SPEC
+<https://github.com/raiden-network/spec/blob/master/messaging.rst>`_.
+
+
--- a/terminology.rst
+++ b/terminology.rst
@@ -75,6 +75,9 @@ Raiden Terminology
    Message
        Any message sent from one Raiden Node to the other.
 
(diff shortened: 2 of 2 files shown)


LefterisJP:
Shouldn't this be part of `transport.rst` ?

pcppcp:
Hm, good question. I've left it in a separate doc because the same message format might be re-used in different transport layers.

hackaugusto:
This document would be better named JSON encoding IMO

palango:
How will we get the sender of a message? Before there was a signature to extract it from.

pcppcp:
There's no specific sender for the message, only for the data transferred. Raiden nodes will i.e. ecrecover the balance proof signature, MS can check BP & reward proof...

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
+#########################
+Raiden transport messages
```
hackaugusto: This document is not describing any message, only the json encoding.

This is best as `JSON encoding`

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
...
+#########################
+
+Transport
+=========
+Messages are exchanged in JSON format using Matrix as a transport layer. Messages can be sent either via broadcast channel or via direct messaging
```
hackaugusto: Why should the transport layer be specified here? The messages are indifferent to the transport used.

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
...
+Message format
+--------------
+The message MUST contain ``body`` field. This field MUST be a JSON object, and it MAY be empty.
+Contents of the field are implementation dependent. Message MAY also contain other fields in the root JSON object, but the application MUST NOT depend on existence of these fields.
+
+
+Simplest valid message is defined as follows
+
+``
+{ 'body': {} }
```
hackaugusto: This is invalid json, use double quotes

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
...
+``
+
+
+Example of a message
+-------------------
+
+``
+{
+    'body': {
```
hackaugusto: this too is invalid json, use double quotes

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
...
+* JSON is easily extensible and extension can be made backward-compatible
+* Message format must not be dependent on the transport used - it should not matter whether the message is sent via Matrix or REST
+* Sender identity verification must not be part of the transport. It should be possible to submit the message over an untrusted channel or delegate.
+
+Fields validation
+-----------------
+* Receiver SHOULD use JSON schema to validate integrity of received message depending on the message type.
+* For messages that contain data later used as a smart contract parameters, receiver MUST check for overflows and conversion errors.
+* Any values to be used on-chain MUST NOT in any circuimstances be floats.
+* For ecrecover-able signatures, Sender identity SHOULD be verified.
```
hackaugusto: Why is this a SHOULD and not a MUST?

json_encoding.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Message format
+--------------
+The message MUST contain ``body`` field. This field MUST be a JSON object, and it MAY be empty.
+Contents of the field are implementation dependent. Message MAY also contain other fields in the root JSON object, but the application MUST NOT depend on existence of these fields.
```
andrevmatos: I don't get why we need a body field. The body value can just be put as the root object/payload. For now, looks like this spec is just to make a `body` field which _separates_/_differentiates_ between data the application MAY care/depend, and data that it must not (metadata in root). I'd rather not force this kind of semanthic or on a specific message layout/structure. We can just make `body` the root object, new [meta]data can simply be added to the body, and the app should only care about the fields it cares. IMHO clear spec on message data itself, ignoring [meta]data which isn't in the spec, is better than adding this kind of unnecessary schema to a higher level spec, which adds complexity to only gain a specific place where one can put metadata it may be sure won't affect the app (why?) and still requires a protocol/agreement on the body format anyway.

pcppcp: `body` has been removed from the message, doc needs an update. The message now looks like exactly how you proposed.

PR #64, opened 2018-04-25
Update MicroRaiden terminology.rst

EvertonMelo:
add micro-Raiden
font: https://raiden.network/faq.html#microraiden


LefterisJP:
This spec is not about microraiden so this is not required.

EvertonMelo:
ok

PR #65, opened 2018-04-25
Update raiden nw terminology.rst

EvertonMelo:
add correct Raiden Network:
font: https://raiden.network/faq.html#raiden

### DIFF
--- a/terminology.rst
+++ b/terminology.rst
@@ -26,7 +26,7 @@ Raiden Terminology
        A network of payment channels for a given Token.
 
    Raiden Network
-       A collection of Token networks.
+       The Raiden Network is an off-chain transfer network for Ethereum ERC20 tokens. 
 
    Transfer
        A movement of tokens from a :term:`Sender` to a :term:`Receiver`.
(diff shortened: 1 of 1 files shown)


LefterisJP:
Thank you for your submission but this is not correct. The raiden network, for the purpose of the specification is a collection of token networks. Succinct and to the point.

EvertonMelo:
then

>is a collection of token networks. Succinct and to the point.

I'll ignore that I put the definition in the PR.

**What is the Raiden Network?**
_The Raiden Network is an off-chain transfer network for Ethereum ERC20 tokens. It provides a fast, scalable, and cheap alternative to on-chain token transfers._ 

so let's go, does raiden fit for all tokens of the ethereum project? can be honest,

so we will ignore the

**What is Raidos?**
_Raidos (or Raiden 2.0, “dos” is spanish for "two") is a proposed sidechain technology to generalize state channels. While the Raiden Network is limited to ERC20 token transfers, Raidos aims to scale Ethereum’s generalized computation capabilities through a network of satellite chains, which can host any smart contract. This technology is similar and complementary to Ethereum's sharding._

see that I have set forth something that you have not even seen.
so in your head is succinct, but in the documentation is generic and misleading.

please be consistent with what is in the documentation, do not create a scope that only exists in your head.
please create objective points do not invent abitrary reasons for not accepting a PR.

PR #68, opened 2018-04-25
Specify secret size

### LINKED ISSUE #60: Use bytes instead of bytes32 for the secret in the registerSecret function
Asses if the `SecretRegistry.registerSecret` function should use `bytes` instead of `bytes32`, to cover all sizes in case we need to change the way the secret is created.

https://github.com/raiden-network/raiden-contracts/blob/81698bb48c94d0a1be556d8b062d316f1e0fc083/raiden_contracts/contracts/SecretRegistry.sol#L20

The `secret` used now in `raiden` repo has a fixes size of 32 bytes
https://github.com/raiden-network/raiden/blob/ba81dd8ba323958e9718a53766457835f11dcd36/raiden/utils/__init__.py#L34

Notes:

- there is no mention in the spec about the size of the `secret`, so one would assume it can be of any size -> either fix the `spec` or do the above change
- difference in gas costs for both solutions must be calculated

LefterisJP:
Fix #60.

@loredanacirstea is that where in the spec you wanted it specified? In all other places it's shown as `bytes32` in the signature of the functions so should be clear.

### DIFF
--- a/terminology.rst
+++ b/terminology.rst
@@ -143,7 +143,7 @@ Raiden Terminology
        Total fees for a Mediated Transfer announced by the Raiden Node doing the Transfer.
 
    Secret
-       A value used as a preimage in a :term:`Hash Time Locked Transfer`.
+       A value used as a preimage in a :term:`Hash Time Locked Transfer`. Its size should be 32 bytes.
 
    Partner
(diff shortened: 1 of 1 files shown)


loredanacirstea:
Exactly, yes.

PR #69, opened 2018-04-25
Proper definition of additional hash

### LINKED ISSUE #62: Add a proper definition of additional_hash
Add a proper definition here:
https://github.com/raiden-network/spec/blob/master/messaging.rst#balance-proof

`Computed from the message. Used for message authentication` is superficial.

Definition should make the following clear:

- when and where is this used
- who calculates it and from what data (mention if this is an unknown or can be customized; give an example)
- who must verify this
- if any 3rd party services must/could use it in some way

Definition details from RC discussions & other docs:

- Because it's not necessarily the hash of the message, it may be zeroed out or just the hash of a subset of the fields. This field is used to add support for authentication without required multiple layers of signatures. Think of the metadata used for the payment, the metadata is unspecified in this spec and the application may want to authenticated it nevertheless.

- Adding salt data to the messages, the goal is to always randomly change `additional_hash` and make it harder for a monitoring service to brute force the hashed balance proof

LefterisJP:
Fix #62


loredanacirstea:
> Hash of additional data used on the application layer. e.g.: payment metadata 

What I would add in `terminology`, aside from the above, is something like:

```
The Raiden Network protocol does not use or enforce the type or content of this additional data.
```

messaging.rst
```diff
@@ -25,25 +25,25 @@ Invariants
...
+|  locksroot            | bytes32     | Root of merkle tree of all pending lock lockhashes                            |
++-----------------------+-------------+-------------------------------------------------------------------------------+
+|  channel_identifier   | uint256     | Channel identifier inside the TokenNetwork contract                           |
++-----------------------+-------------+-------------------------------------------------------------------------------+
+| token_network_address | address     | Address of the TokenNetwork contract                                          |
++-----------------------+-------------+-------------------------------------------------------------------------------+
+| chain_id              | uint256     | Chain identifier as defined in EIP155                                         |
++-----------------------+-------------+-------------------------------------------------------------------------------+
+|  additional_hash      | bytes32     | Hash of additional data used on the application layer. e.g.: payment metadata |
```
palango: Comma before *e.g.*

LefterisJP: Comman instead of the full stop you mean?

palango:
I like the addition from Loredana, but I would add it to the *Invariants* list

LefterisJP:
@palango 

> but I would add it to the Invariants list

Why there? This should be part of what additional_hash is.

palango:
Because that's where the balance proof details are explained. I would just add an additional point there:
- `additional_hash` is the hash of arbitrary application layer data. The type or content of this additional data is up to the application.

messaging.rst
palango: yes

palango:
Superseeded by #72

LefterisJP:
@palango I would appreciate you pinging me before closing my PR.

Also if you want to add additional commits to someone else's PR you can simply push to that branch and they are added to the PR. No need to open a different PR.

palango:
@LefterisJP It was only a small change. Will do next time.

I couldn't push to your branch, github said I don't have permissions.

LefterisJP:
> I couldn't push to your branch, github said I don't have permissions.

You probably did it wrong then. Others have succesfully added commits to my PRs in the past, and adding commits is on by default.

PR #72, opened 2018-04-27
Define additional hash

palango:
Fix a small typo in #69


palango:
Superseedes #69 
Fixes #62

PR #93, opened 2018-06-27
Protocol diagrams - messages, on-chain channel lifecycle

loredanacirstea:
- diagram sources & images (see added Readme.md for how to use the source to build the diagrams)
- linked diagrams in the spec


messaging.rst
```diff
@@ -349,6 +349,7 @@ For the simplest Alice - Bob example:
 - Alice creates a new transfer with:
     * transferred_amount = ``current_value``
     * lock = ``Lock(n, hash(secret), expiration)``
+    * locked_amount = ``updated value containing the lock amount``
     * locksroot = ``updated value containing  the lock``
```
palango: Not related to your PR, but there's a space to much between *containing* and *the*.

loredanacirstea:
Fixed the typo, thank you.

onetom:
I was also wondering why are the diagram source filenames don't match the image filenames?

It would be easier to script the image regeneration if the names would match.
It would even make it unnecessary to commit the images, the same way as the `_build/` directory
is not committed, since it's just an output of `sphinx`.

loredanacirstea:
@onetom , I should have used SVG's indeed. No reason not too, it wouldn't have been hard at that time. The filenames can be renamed, they were created manually.

I personally use an editor I created, that handles Markdown and a couple of diagram libraries & has live preview + I also have a server for it. So, for me personally it is about convenience of creating the diagrams, but it would be good to also automate the file generation, yes. 

Otherwise, we have not discussed diagrams as a documentation standard for Raiden Network. I am open to anything that is easy to use. So, maybe @LefterisJP has an opinion on this.

onetom:
I've just pushed some updates to my fork and it can render both the sequence and the state diagrams automatically.

I've turned all of them into SVG too.

I had to change the `&` sign, because the nomnoml cli emitted invalid SVG because of that. (didn't do HTML entity encoding on the text)

onetom:
@loredanacirstea are you the author of https://typora.io ?
because the editor you described sounds very much like that.

I love that editor and I wish it would be available under Linux too.
I like the fact that it can render various diagram formats, but those
diagramming libraries are not very mature and they don't have
integration with editors and IDEs, that's why I settled on PlantUML,
despite being slower and a lot heavier software package.

loredanacirstea:
@onetom , ah no. I did not know about https://typora.io! It is indeed similar in features, but way less polished and nice (made it 2 years ago, never improved it, still functional). I will check them out, thank you.

Issue #86, opened 2018-06-14
Add invariants for how the settlement amounts must be computed


loredanacirstea:
Check the conclusions from https://github.com/raiden-network/raiden-contracts/pull/123, and those obtained after solving https://github.com/raiden-network/raiden-contracts/issues/127 and clearly define the protocol used to compute the channel settlement amounts for each participant:
- all the values needed and what they stand for
- value constraints/invariants and who enforces them
- algorithm used to compute the amounts (not the Solidity implementation, we are not interested in the Solidity quirks - overflows and underflows, but in what the actual logic is when calculating the settlement balances)

Link all settlement computation issues & PRs as comments in this issue, to keep track of changes.

loredanacirstea:
I am making a settlement algorithm analysis & attack vectors here: https://github.com/raiden-network/raiden-contracts/issues/188. It also contains some of the things that are supposed to be in the spec.

jomuel:
Ok, I spent yesterday recovering & documenting the algorithm from the contract code. I'll see what needs to be rewritten.

loredanacirstea:
@jomuel , that is perfect. Then please check that what I've written is correct. If you think the settlement algorithm should be different or you see differences between what you understood and what I've written, please document them.

loredanacirstea:
I updated the description to make it more clear that the issue is not for describing the settlement algorithm as implemented in Solidity, but for explaining the settlement logic and value constraints.

This was unclear previously.

Issue #85, opened 2018-06-11
Update balance proof spec


loredanacirstea:
Now, we have the balance proof that the smart contracts need to receive in `messages.py`. As pointed out by @hackaugusto , this data structure is not used per se by the messages. They can keep their current structure - with all balance data, computing the `balance_hash` when needed.

TODO: 
- move the balance proof definition in the smart contract spec from here https://github.com/raiden-network/spec/blob/master/messaging.rst#balance-proof
- add the actual data structure used by the messages, containing the actual data.

loredanacirstea:
Fixed by https://github.com/raiden-network/spec/pull/96

PR #96, opened 2018-07-13
Remaining issues for the red eyes milestone (#85, #86)

jomuel:
As for the balance computing algorithm, I just tried to describe it as it is implemented now. I will check against the new analysis of the algorithm in the raiden-contracts repo later.


messaging.rst
```diff
@@ -162,6 +43,8 @@ Fields
 Messages
 ========
 
+.. _direct-transfer-message:
+
 Direct Transfer
```
loredanacirstea: @LefterisJP ,
Are we keeping the direct transfer in the specs? From what I know, we do not have it anymore.

messaging.rst
```diff
@@ -9,125 +9,6 @@ This is the specification document for the messages used in the Raiden protocol.
...
-+------------------------+------------+--------------------------------------------------------------------------------+
-|  channel_identifier    | uint256    | Channel identifier inside the TokenNetwork contract                            |
-+------------------------+------------+--------------------------------------------------------------------------------+
-| token_network_address  | address    | Address of the TokenNetwork contract                                           |
-+------------------------+------------+--------------------------------------------------------------------------------+
-| chain_id               | uint256    | Chain identifier as defined in EIP155                                          |
-+------------------------+------------+--------------------------------------------------------------------------------+
-|  signature             | bytes      | Elliptic Curve 256k1 signature on the above data                               |
-+------------------------+------------+--------------------------------------------------------------------------------+
-
```
loredanacirstea: In the `Data Structures` section I would leave the following data structures:
- Balance Proof - defined as https://github.com/raiden-network/raiden/blob/master/raiden/transfer/state.py#L459, in a similar format as defined initially (see 
 also https://github.com/raiden-network/spec/blob/7134b3fce660c85b132f2e6e3f182715cd45a32e/messaging.rst#balance-proof), but with up-to-date fields.

Then, reference it the `Messages` as needed, without repeating the fields.
Balance proof exists as a concept both in the smart contracts and the Raiden client.

messaging.rst
```diff
@@ -175,39 +58,58 @@ Invariants
...
+|  locked_amount       | integer       | Amount to be transferred including fees                     |
++----------------------+---------------+-------------------------------------------------------------+
+|  fee                 | integer       | Remaining amount to be used as transfer fees                |
++----------------------+---------------+-------------------------------------------------------------+
+|  target              | address       | Final target for this transfer                              |
++----------------------+---------------+-------------------------------------------------------------+
+|  secrethash          | hash          | Hash of the secret that will unlock the transfer            |
++----------------------+---------------+-------------------------------------------------------------+
 
+Fees are always payable by the initiator. The intermediary node receiving this can deduct its own fee and lower ``fee`` to the remaining fee. It is not incentivised to take a too high fee, as this would render completion of the transfer unlikely.
```
loredanacirstea: I was not aware of any current `fee` discussions for the spec & implementation, so I cannot review this.
@LefterisJP ?

messaging.rst
```diff
@@ -235,7 +137,7 @@ Fields
 Secret Reveal
 -------------
 
-Message used by the nodes to inform others that the :term:`secret` is known. Used to request an updated :term:`balance proof` with the :term:`transferred amount` increased and the lock removed.
+Message used by the nodes to inform others that the :term:`secret` is known. Used to request data to update the :term:`balance proof` with the :term:`transferred amount` increased and the lock removed.
```
loredanacirstea: In my opinion, the initial version is better.

The node does request `an updated balance proof`. Balance proof is a concept and it is represented by https://github.com/raiden-network/raiden/blob/master/raiden/transfer/state.py#L459

At least, this is the terminology we have been using until now.

messaging.rst
```diff
@@ -175,39 +58,58 @@ Invariants
...
 Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the corresponding merkle tree of the sending participant node.
 
 
 Invariants
 ^^^^^^^^^^
 
-- The :term:`balance proof` locksroot must be equal to the previous valid merkle tree with the lock provided in the messaged added into it.
-- The transfer is valid only if the lock amount is smaller than the sender's :term:`capacity`.
+- The :term:`locksroot` must be equal to the previous valid merkle tree with the lock provided in the messaged added into it.
+- The transfer is valid only if the ``locked_amount`` is smaller than the sender's :term:`capacity`.
```
loredanacirstea: Same for other occurrences - I would use the balance proof concept and keep the initial version.

smart_contracts.rst
```diff
@@ -337,6 +461,35 @@ Unlocks all pending transfers by providing the entire merkle tree of pending tra
     ``unlock`` must be called after ``settleChannel`` because it needs the ``locksroot`` from the latest :term:`balance proof` in order to guarantee that all locks have either been unlocked or have expired.
 
 
+.. _settle-computation:
+
+Computation of balances
+~~~~~~~~~~~~~~~~~~~~~~~
+
+There is an elaborate algorithm to safely compute the balances, avoiding simple errors like over- and underflows as well as vulnerabilities for cheating.
+A possible attempt at cheating would be to present an outdated balance proof before an unlocking event, so that the locked amount would be assessed too high.
```
loredanacirstea: I would just resume to the settlement constraints and leave the analysis about cheating out of here for now. I don't think the security analysis of the settlement algorithm should be in the spec.
Also, `A possible attempt at cheating` might be interpreted wrongly.

smart_contracts.rst
```diff
@@ -337,6 +461,35 @@ Unlocks all pending transfers by providing the entire merkle tree of pending tra
...
+
+This value, capped by the total available deposit, is assigned to participant 1 as their balance: ``B1 = min(TAD, MR1)``
+
+The balance of participant 2 is then deducted by the invariant stated above: ``B2 = TAD - B1``
+
+The **actually locked amounts** will be the locked amounts capped by the respective partners' balance: ``La1 = min(L1, B2); La2 = min(L2, B1)``
+
+Finally, the **actually transferred amounts** will be balances minus actually locked amounts: ``Ta1 = B1 - La2; Ta2 = B2 - La1``
+
+If ``Ta1 > 0`` or ``Ta2 > 0``, they are transferred in the settlement of the channel.
```
loredanacirstea: As said above, I would concentrate on the constraints here, rather than the algorithm. The specs should be guidelines for anyone wanting to implement the protocol. The settlement algorithm could possibly be implemented in other ways too.
I would rather have this in another PR. And I will actually make one with what I wrote in https://github.com/raiden-network/raiden-contracts/issues/188#issuecomment-404752095 if you don't mind.

loredanacirstea: I did the PR here: https://github.com/raiden-network/spec/pull/99

messaging.rst
jomuel: To move the balance proof definition to the smart contracts session was an explicit todo in issue #85.

messaging.rst
jomuel: "Balance proof" is introduced to the specs as the name of a data structure. In that sense, the node does not request a balance proof, but the necessary data to construct the updated balance proof. As I understand issue #85, it is about being accurate about this? (Although we could change the formulation here to point out the balance proof still comes from the partner, something like: "Used to request the missing data to construct the partner's updated balance proof")

messaging.rst
loredanacirstea: Yes, the issue description was not clear enough.
The problem was that the balance proof definition from `messages.py` is not correct from the Raiden client perspective, but it is correct from the smart contracts perspective -> should be moved to the smart contracts.

As mentioned in the issue:
> add the actual data structure used by the messages

For `messages.py` we still need to define this Balance Proof data structure, but with the correct components from the Raiden client perspective - e.g. have `transferred`, `locked_amount`, `locksroot` instead of the contract's `balance_hash`.
So, I failed to add a reference to what needs to be used in the spec here, which is: https://github.com/raiden-network/raiden/blob/1ca9e251fe33564a430c06d9a9b49614daa7b916/raiden/transfer/state.py#L466-L482

Now, this was my take on it. Maybe someone else has another opinion on the specs.

smart_contracts.rst
jomuel: I removed the whole commit related to issue #86 from this pr since pr #99 covers it.

messaging.rst
jomuel: Ok, I'll change the `Messages` section accordingly. I'll also revert the two single-line changes you commented below; if it is not about individual data structures for the different messages the distinction between balance proof and data to update the balance proof is not so important.

messaging.rst
loredanacirstea: I don't think your version is incorrect. I personally preferred to maintain the balance proof concept because I think this is more important than the actual implementation (transferred amount, locked, locksroot). In the sense that the balance proof can (maybe) be implemented in other ways by people, but serves the same scope.

This being said, maybe some input from @LefterisJP or @hackaugusto might be better regarding this.

messaging.rst
loredanacirstea: Update: we still have it and it works. I was not aware.

smart_contracts.rst
```diff
@@ -40,6 +40,128 @@ Project Requirements
...
+
+Fields
+^^^^^^
+
++------------------------+------------+--------------------------------------------------------------------------------+
+| Field Name             | Field Type |  Description                                                                   |
++========================+============+================================================================================+
+|  balance_hash          | bytes32    | Balance data hash                                                              |
++------------------------+------------+--------------------------------------------------------------------------------+
+|  nonce                 | uint64     | Strictly monotonic value used to order transfers. The nonce starts at 1        |
```
loredanacirstea: Also, `nonce` is now `uint256` - https://github.com/raiden-network/raiden-contracts/blob/dd0b258d79622211a8605012038d39ef710b0f67/raiden_contracts/contracts/TokenNetwork.sol#L83, we need to also fix this in the spec.

jomuel:
The changes I pushed now address all comments except for the line about fees. There may be a merge conflict with PR #103 but there are no contentual conflicts.

PR #103, opened 2018-08-15
Missing constraints for the direct transfer

hackaugusto:
The direct transfer must only change the transferred amount, the merkle
tree must stay intact, therefore the locked_amount and the locks_root
must be not change in respect to the previous message. Additionally the
message must be for a valid channel, meaning it must be tied to a
specific channel, from a specific token network, for a specific
blockchain, and the channel must be open.

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -170,7 +170,13 @@ A non cancellable, non expirable payment.
 Invariants
 ^^^^^^^^^^
 
-- Only valid if the :term:`transferred amount` is larger than the previous value and it increased by an amount smaller than the participant's current :term:`capacity`.
+- Only valid if all of the following hold:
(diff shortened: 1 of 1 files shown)


hackaugusto:
@jomuel You may want to have a look at this PR, since it can conflict with PR #96

hackaugusto:
Additionally, I did *not* check if the terms used are actually defined. I can do this as long as we agree on how to integrate these changes.

LefterisJP:
> Additionally, I did not check if the terms used are actually defined. I can do this as long as we agree on how to integrate these changes.

Doesn't compiling the docs in rst generate warnings for terms which are not defined?

loredanacirstea:
Yes, it generates warnings.
I've defined the missing terms myself in https://github.com/raiden-network/spec/pull/105.

LefterisJP:
Okay I remembered well then. Thank you. Still going through my notifications from secondary repos :)

PR #104, opened 2018-08-15
Missing constraints for mediated transfers

hackaugusto:
The mediated transfer must only change the the merkle tree, the
transferred amount must stay intact, therefore the locked_amount and the
locks_root must change to include the new lock in respect to the
previous message. Additionally the message must be for a valid channel,
meaning it must be tied to a specific channel, from a specific token
network, for a specific blockchain, and the channel must be open.

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -110,8 +110,15 @@ Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is be
 Invariants
 ^^^^^^^^^^
 
-- The :term:`balance proof` locksroot must be equal to the previous valid merkle tree with the lock provided in the messaged added into it.
-- The transfer is valid only if the `locked_amount` is smaller than the sender's :term:`capacity`.
+- Only valid if all the following hold:
+    - There is a channel which matches the given :term:`chain_id`, :term:`token_network`, and :term:`channel_identifier`.
+    - The corresponding channel is in the open state.
+    - The :term:`nonce` is increased by one in respect to the previous :term:`balance proof`
+    - The :term:`locksroot` must change, the new value must be equal to the root of a new tree, which has all the previous locks plus the lock provided in the message.
+    - The :term:`locked_amount` must increase, the new value is equal to the old value plus the lock's amount.
+    - The lock's amount must be smaller then the participant's :term:`capacity`.
+    - The lock expiration must be greater than the current block number. 
+    - The :term:`transferred amount` must not change.
 
 Fields
 ^^^^^^
--- a/messaging.rst
+++ b/messaging.rst
@@ -105,8 +105,6 @@ Locked Transfer
 
 Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the corresponding merkle tree of the sending participant node.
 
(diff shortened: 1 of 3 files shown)


smart_contracts.rst
```diff
@@ -694,7 +694,7 @@ Must be enforced by the Raiden Client:
 .. Note::
     Any two consecutive balance proofs for ``P1``, named ``BP1k`` and ``BP1t`` were `k < t`,  must respect the following constraints:
 
-    1. A :term:`Direct Transfer` or a :term:`Mediated Transfer` with ``value`` tokens was finalized, therefore ``T1t == T1k + value`` and ``L1t == L1k``.
+    1. A :term:`Direct Transfer` or a :term:`Locked Transfer` with ``value`` tokens was finalized, therefore ``T1t == T1k + value`` and ``L1t == L1k``.
```
loredanacirstea: What I wrote here contradicts the explanation you wrote above for `Locked Transfer`. I am here referring to the full, finalized transfer. You were defining only the `Locked Transfer` message.

I propose to keep both `Locked Transfer Message` concept and `Mediated Transfer`, where:
`Mediated Transfer` = the finalized transfer, composed of multiple Raiden client messages (`LockedTransfer`, `RevealSecret`, `Unlock` ...)
`Locked Transfer (Message)` = the first message that starts the mediated transfer, as defined above by you.

messaging.rst
```diff
@@ -100,26 +100,31 @@ Fields
 |  balance_proof       | BalanceProof  | Balance proof for this transfer                            |
 +----------------------+---------------+------------------------------------------------------------+
 
-Mediated Transfer
+Locked Transfer
 -----------------
 
 Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the corresponding merkle tree of the sending participant node.
```
loredanacirstea: Is the `Locked Transfer` message a `transfer`? 
`transfer = A movement of tokens from a Sender to a Receiver.` from the terminology. It does not say anything about the finality - wether it is `non-refundable` or `non-cancellable`

hackaugusto: > Is the Locked Transfer message a transfer?

If we define the transfer as the *movement of tokens*, then the `Locked Transfer` message alone does not qualify. A transfer would be initialized by the `LockedTransfer` message and finalized by the `Unlock` message.

>  It does not say anything about the finality - wether it is non-refundable or non-cancellable

The transfer initiated with locked with the message `LockedTransfer` is cancellable, and refundable. 

The transfer can be cancelled if the initiator doesn't send the first secret reveal message, and wait for it to expire.

The transfer may be refunded to a mediator by another, if another message with a lock with the same amount and secrethash is sent in reverse.

hackaugusto: If we use transfer to the movement of tokens, then we would have two types of transfers:

- Direct transfers, implemented by a single direct transfer message. The transfer is non-expirable and non-refundable.
- Locked Transfers. The transfer is implemented by the messages LockedTransfer, RevealSecret, SecretRequest, Unlock. The *transfer* is expirable, cancellable. The message LockedTransfer is refundable.

smart_contracts.rst
hackaugusto: Sounds good :)

loredanacirstea: Yes, on the above line it should be `Mediated Transfer` , instead of `Locked Transfer`.

PR #105, opened 2018-08-17
Spec fixes, code style

loredanacirstea:


smart_contracts.rst
```diff
@@ -694,14 +694,14 @@ Must be enforced by the Raiden Client:
 .. Note::
     Any two consecutive balance proofs for ``P1``, named ``BP1k`` and ``BP1t`` were `k < t`,  must respect the following constraints:
 
-    1. A :term:`Direct Transfer` or a :term:`Locked Transfer` with ``value`` tokens was finalized, therefore ``T1t == T1k + value`` and ``L1t == L1k``.
-    2. A :term:`HTL` with ```value`` was sent, but not yet finalized, therefore ``T1t == T1k`` and ``L1t == L1k + value``.
+    1. A :term:`Direct Transfer` or a succesfull :term:`HTL Transfer` with ``value`` tokens was finalized, therefore ``T1t == T1k + value`` and ``L1t == L1k``.
```
hackaugusto: Wasn't the plan to use `Mediated Transfer` for these cases?

loredanacirstea: Yes, in my comment from here: https://github.com/raiden-network/spec/pull/104/files#r210553993 I was saying about using `Mediated Transfer`. But I realized that a `Mediated Transfer` might be confusing here, because it can be composed from multiple `HTL` transfers.

The goal of the above explanation is just to clear up that `T` (transferred amount) is the one increased when a transfer is finalized. The `HTL` concept is simpler and does the work in my opinion.

PR #107, opened 2018-08-20
fix typo in link in index.rst

BOR4:
raiden.nework leads nowhere.


BOR4:
@loredanacirstea I don't know why it does not work. I signed it when I made PR.

image

PR #109, opened 2018-08-20
fix url on homepage

BOR4:


BOR4:
@loredanacirstea sorry for making a mess in repository. I had to set my email locally on my laptop and unfortunately old commits could not be affected by a change.

Looks like it is ok now

loredanacirstea:
@BOR4 no problem, thank you!

PR #119, opened 2018-08-30
Fix explanation for settlement overflow

loredanacirstea:


smart_contracts.rst
```diff
@@ -760,6 +760,13 @@ Also, the amount that a participant can receive cannot be bigger than the total
 
 ``(6 R)`` is deduced from ``(5 R)``. It is needed by the Raiden client in order to not allow a participant to :ref:`withdraw <withdraw-channel>` more tokens from the on-chain channel deposit than he is entitled to.
 
+::
+
+    (7 R) T2 + L2 - T1 - L1 <= D2
+
+``T2 + L2 - T1 - L1`` is the netted total transferred amount from ``P2`` to ``P1``. This amount cannot be bigger than ``P2``'s deposit -> a participant cannot transfer more tokens than what he has in the channel, during the lifecycle of a channel.
+This ``MUST`` be ensured by the Raiden client.
```
loredanacirstea: So, I think `(7 R) T2 + L2 - T1 - L1 <= D2` is actually what the Raiden client `MUST` enforce.
However, the Raiden client only enforces `(5 R) AB1 = D1 - W1 + T2 - T1 - L1; AB1 >= 0, AB1 <= TAD` as far as I know.
So, we might want to prove that `(5 R)` ensures `(7 R)`

smart_contracts.rst
```diff
@@ -760,6 +760,13 @@ Also, the amount that a participant can receive cannot be bigger than the total
 
 ``(6 R)`` is deduced from ``(5 R)``. It is needed by the Raiden client in order to not allow a participant to :ref:`withdraw <withdraw-channel>` more tokens from the on-chain channel deposit than he is entitled to.
 
+::
+
+    (7 R) T2 + L2 - T1 - L1 <= D2
+
+``T2 + L2 - T1 - L1`` is the netted total transferred amount from ``P2`` to ``P1``. This amount cannot be bigger than ``P2``'s deposit -> a participant cannot transfer more tokens than what he has in the channel, during the lifecycle of a channel.
```
hackaugusto: What about mentioning the lower bound? This value can be negative but not smaller than D1. (This is a corollary that must be enforce by P2)

loredanacirstea: Correct, added.

smart_contracts.rst
```diff
@@ -762,9 +762,10 @@ Also, the amount that a participant can receive cannot be bigger than the total
 
 ::
 
-    (7 R) T2 + L2 - T1 - L1 <= D2
+    (7 R) -D1 <= T2 + L2 - T1 - L1 <= D2
```
hackaugusto: btw, this is missing the withdraws:

`-(D1 - W1) <= T2 + L2 - T1 - L1 <= D2 - W2`

loredanacirstea: Are you sure? `T + L` are the total transferred amounts. The netted total transferred amounts should be bounded by the channel deposit. 
The withdraws are then taken from the available balance, but we do not subtract them from the total transferred amounts.

hackaugusto: > The netted total transferred amounts should be bounded by the channel deposit.

Nope, they are bound by the participant's capacity, which is D - W.

You cannot have a net value that is larger then your available deposit on chain, consider:

P2 Deposit 10
P2 Withdraws 10
If P1 enforce just `T2 + L2 - T1 - L1 <= D2` it means P2 can make transfer up to 10 tokens, which is wrong.

Additionally, since you're talking about the off-chain client, `D2 - W2` is **not** sufficient, since `W2` is defined as the current on-chain withdraw. The client must enforce this against the **off**-chain withdraw, which is different from the on-chain value.

loredanacirstea: I think you did not consider `P1`'s transferred amounts. `P2` can transfer up to `10` tokens if `P1` owes him `10` tokens.

E.g.:

```
D1 = 10
W1 = 0
T1 = 4
L1 = 0


D2 = 10
W2 = 5
T2 = 10
L2 = 0
```

```sh
T2 + L2 - T1 - L1 = 10 + 0 - 4 - 0 = 6
# The following is true:
# - D1 <= 6 <= D2
# - 10 <= 6 <= 10

# The following is not true:
# -(D1 - W1) <= 6 <= D2 - W2
# - (10 - 0) <= 6 <= 10 - 5
# -10 <= 6 <= 5
# 6 <= 5 not true
```

smart_contracts.rst
```diff
@@ -760,6 +760,14 @@ Also, the amount that a participant can receive cannot be bigger than the total
...
 ``(6 R)`` is deduced from ``(5 R)``. It is needed by the Raiden client in order to not allow a participant to :ref:`withdraw <withdraw-channel>` more tokens from the on-chain channel deposit than he is entitled to.
 
+::
+
+    (7 R) -(D1 - W1) <= T2 + L2 - T1 - L1 <= D2 - W2
+
+``T2 + L2 - T1 - L1`` is the netted total transferred amount from ``P2`` to ``P1``. This amount cannot be bigger than ``P2``'s **available** deposit. We enforce that a participant cannot transfer more tokens than what he has in the channel, during the lifecycle of a channel.
+This amount cannot be smaller than the negative value of ``P1``'s **available** deposit ``- (D1 - W1)``. This can also be deducted from the corresponding ``T1 + L1 - T2 - L2 <= D1 - W1``
+The Raiden client ``MUST`` ensure this. However, it must use up-to-date values for ``D2`` and ``W2`` (e.g. Raiden node might have sent an on-chain transaction to withdraw tokens; this is not mined yet, therefore it does not reflect in the contract yet. The Raiden client will use the off-chain ``W2`` value.)
+
```
loredanacirstea: @hackaugusto , I added the off-chain values mention.

PR #124, opened 2018-09-19
smart_contracts.rst with terms not assuming participant is human

mat7ias:
Neutral terms would suit better in these sections

### DIFF
--- a/smart_contracts.rst
+++ b/smart_contracts.rst
@@ -13,8 +13,8 @@ General Requirements
 Secure
 ------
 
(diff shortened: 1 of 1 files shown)


mat7ias:
@loredanacirstea When the first noun is an epicene, 'they' and 'their' is the most suited to refer to that person (the reader of the specs in this case). It shouldn't bring confusion within the context of the specs->reader. I could make another suggestion but it'd be suboptimal imo.

pirapira:
I like most of the changes except the additional occurrences of `they` referring to singlular nouns.

While this PR was open it has become out of sync from the repo.  So I'm closing this.

I recommend smaller PRs going forward (either doing one simple kind of changes, or touching only several lines).

PR #131, opened 2018-11-16
Distinguish two kinds of balance proofs

### LINKED ISSUE #129: Similar specifications of "Balance Proof" in message spec and contract spec
The [messages specification](https://github.com/raiden-network/spec/blob/cddb561d925264c03fbecb7c9e861249c9a874c1/messaging.rst#balance-proof) and the [smart contracts specification](https://github.com/raiden-network/spec/blob/23a74c986ccab673673610469347a1837f23f062/smart_contracts.rst#balance-proof) both specify what "Balance Proof" is.  The two versions differ.

I'm guessing the two specifications define different structures for different contexts (but I'm not totally sure).  An explicit note would be an improvement.

pirapira:
The balance proofs have different formats for offchain messasges and
onchain calls into smart contracts.  This commit gives different names
for these two kinds of balance proofs: onchain balance proofs and
offchain balance proofs.

Fixes #129.


messaging.rst
```diff
@@ -50,7 +54,7 @@ Fields
 Merkle Tree
 -----------
 
-A binary tree composed of the hash of the locks. The root of the tree is the value used in the :ref:`balance proofs <balance-proof-message>`. The tree is changed by the ``LockedTransfer``, ``RemoveExpiredLock`` and ``Unlock`` message types.
+A binary tree composed of the hash of the locks. The root of the tree is the value used in the :ref:`onchain balance proofs <balance-proof-onchain>`. The tree is changed by the ``LockedTransfer``, ``RemoveExpiredLock`` and ``Unlock`` message types.
```
loredanacirstea: I would mainly reference the `offchain` balance proof here, because this is the one used in the client in conjunction with the merkle tree. The reference to the `onchain` one (in the offchain balance proof definition) should be enough to make the connection with the contracts.

pirapira: I'll change the link to terminology::Balance Proofs.

pirapira: Done.

pirapira:
I did something different from what @loredanacirstea said.  Is this good?

loredanacirstea:
Yes, it's even better.

Issue #133, opened 2018-11-16
Remove sender field from messages with signatures


pirapira:
As @hackaugusto [says](https://github.com/raiden-network/spec/issues/130#issuecomment-439431552), since message senders should be computed from the signature, the sender fields should be removed from messages with signatures.

I've already observed that sender field is not necessary: https://github.com/raiden-network/spec/issues/130#issue-381589508

* [x] remove the sender field from the spec
* [ ] remove the parsing of `sender` field from the implementation, when it deals with signed messages (this can be ticked when a issue is created in the impl repository).

hackaugusto:
As for the parsing, the client and the smart contracts already behave correctly, not sure about the other services. @palango do you know?

pirapira:
I'll have a look into Monitoring Service, and show @palango what I find.

palango:
@pirapira It uses the same foundational code in the raiden-libs, so it should be the same.

pirapira:
I started from the Monitoring Service, and found the `raiden-lib` code that @palango mentioned.  https://github.com/pirapira/raiden-libs/blob/c20a9aa1e84bb4afa36cc5e5e50693ac4215fa84/raiden_libs/messages/balance_proof.py#L145

pirapira:
@loredanacirstea showed me [a class with `sender` field](https://github.com/raiden-network/raiden/blob/fc6028e14c15448c9f3a08b10a5958c0863493d2/raiden/transfer/state.py#L811).  She tells me objects of this class `BalanceProofSignedState` will be sent over the network.  Reopening the issue to clear this out.

LefterisJP:
As I mentioned in comments [1](https://github.com/raiden-network/spec/pull/135#issuecomment-442041772) and [2](https://github.com/raiden-network/spec/pull/135#issuecomment-442042478) in the spec repo:

In the client, `BalanceProofUnsignedState` is not what is sent to or received from the wire. It's just a nice construct we use to keep the balance proof along with its sender inside the state machine and the client in general.

What goes in the wire is the [packed balance proof](https://github.com/raiden-network/raiden/blob/fc6028e14c15448c9f3a08b10a5958c0863493d2/raiden/transfer/balance_proof.py#L15) which does not contain a `sender.

pirapira:
@LefterisJP I had a look too.  Thank you.

@loredanacirstea if you agree, please close this issue.

loredanacirstea:
Yes, I agree.

PR #135, opened 2018-11-26
[figuring out what's on wire] Update fields of OffchainBalanceProof

pirapira:
As I see in [the implementation](https://github.com/raiden-network/raiden/blob/3a4f3105722d64cdd7aa3abdf9526d8f0cb17b20/raiden/transfer/balance_proof.py#L7),
BalanceProof message should contain seven fields except the signature
and the message type.

This commit tries to update the spec following the implementation.

~~[**need help**] The description of `additoinal_hash` has to be filled, but I don't know what it is.~~ Thanks @LefterisJP and @loredanacirstea for jumping in.


messaging.rst
```diff
@@ -31,22 +31,18 @@ Fields
 +==========================+============+================================================================================+
 |  nonce                   | uint256    | Strictly monotonic value used to order transfers. The nonce starts at 1        |
 +--------------------------+------------+--------------------------------------------------------------------------------+
-|  transferred_amount      | uint256    | Total transferred amount in the history of the channel (monotonic value)       |
-+--------------------------+------------+--------------------------------------------------------------------------------+
-|  locked_amount           | uint256    | Current locked amount                                                          |
-+--------------------------+------------+--------------------------------------------------------------------------------+
```
pirapira: I shouldn't have removed these two fields above.  I was again confused about the onchain/offchain formats.  Will fix.

LefterisJP:
> [need help] The description of additoinal_hash has to be filled, but I don't know what it is.

@pirapira 

As per the [terminology](https://github.com/raiden-network/spec/blob/master/terminology.rst) document:

> Hash of additional data used on the application layer. This can for example be some form of payment metadata. The Raiden Network protocol does not use or enforce the type or content of this additional data.

messaging.rst
pirapira: I've put those fields in again.

pirapira:
I think this PR is ready.

LefterisJP:
As @loredanacirstea pointed out above the `BalanceProofSignedState` is the way we keep a signed balance proof in the client. The previous spec was following the order of the constructor arguments.

To see how we send and receive balance proofs from the wire have a look a [pack_balance_proof](https://github.com/raiden-network/raiden/blob/fc6028e14c15448c9f3a08b10a5958c0863493d2/raiden/transfer/balance_proof.py#L6).

The `balance_hash` is a hash of the `transfered_amount`, `locked_amount` and `locksroot` performed [here](https://github.com/raiden-network/raiden/blob/fc6028e14c15448c9f3a08b10a5958c0863493d2/raiden/transfer/utils.py#L50).

pirapira:
I'll add a link to https://github.com/raiden-network/spec/blob/2a302915293778285323240cb6b63065fb475cbf/smart_contracts.rst#balance-data-hash .

LefterisJP:
> Also, we should fix the sender field - we probably need to remove it from the BalanceProofSignedState

@loredanacirstea Why? This is just a construct the client uses to easily keep the balance proof's sender and compare it in the state machine. It's populated during decoding of the message by doing an `ecrecover` on the signature.

pirapira:
@LefterisJP @loredanacirstea "removing the sender discussion" >> https://github.com/raiden-network/spec/issues/133 perhaps.

loredanacirstea:
@LefterisJP , there was a discussion and a decision for it to be removed: https://github.com/raiden-network/spec/issues/133, I assume it was only removed from the `BalanceProofUnsignedState `.

pirapira:
In the end, offchain balance proofs appear only in the `Secret` messages.  So the actual fields can be seen [Secret class's pack() method](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L645).

pirapira:
I don't see `additional_hash` in the `Secret` message (which should contain offchain balance proofs).  https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L645

Maybe `additional_hash` is not supposed to be in the offchain balance proofs?  @LefterisJP

pirapira:
@LefterisJP 

> To see how we send and receive balance proofs from the wire have a look at pack_balance_proof.

I'm still confused.  `pack_balance_proof` packs `balance_hash` but not `transferred_amount` or `locked_amount`.  @loredanacirstea told me that the Raiden protocol messages should contain `transferred_amount` and `locked_amount`.

LefterisJP:
@pirapira I wrote some things in the  [other](https://github.com/raiden-network/spec/issues/133#issuecomment-442044297) issue. If you read what I wrote there you will see that the actual balance proof that goes in the wire (iow the messages we send offchain to our partners) is the signature of the packed data.

To see which messages that is check all messages.py messages which inherit from [SignedMessage](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L219).

The data that is signed can be seen [here](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L227) for all SignedMessages which don't contain balance proofs.

The messages that contain balance proofs are the ones inheriting from [EnvelopeMessage](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L269) and the data they sign is [here](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L310).

> To see how we send and receive balance proofs from the wire have a look at pack_balance_proof.
> I'm still confused. pack_balance_proof packs balance_hash but not transferred_amount or locked_amount. @loredanacirstea told me that the Raiden protocol messages should contain transferred_amount and locked_amount.

Yeah she is correct. I think my original phrasing could lead to misunderstanding. I should have said. That `pack_balance_proof` packs the data that is to be signed. Additional hash is part of them. Then they get signed and they become the [signature](https://github.com/raiden-network/raiden/blob/133619dff356b0d8bea054ed156dbabcf3ab98f8/raiden/messages.py#L865) part of what goes in the wire.

To see what goes in the wire for each message check its `pack()` function.

pirapira:
@LefterisJP yes,  @loredanacirstea gave me a tour around that overriding `_data_to_sign`.  Now I know what's on wire.

Then @konradkonrad asked an interesting question: https://github.com/raiden-network/raiden/issues/3082.  Also https://github.com/raiden-network/raiden/issues/3083 came up.

pirapira:
This thread turned into an investigation thread.  Now I'm starting the patch almost from the scratch, so closing the PR.

PR #136, opened 2018-11-28
Improve description of offchain balance proofs

pirapira:
Some improvements on the description of offchain balance proofs.

* nonces number messages, not transfers
* Explicitly say it's a set of fields.
* Move the signature section after the table of fields.
* Explain that the signature is the same for onchain and offchain balance proofs.
* Explain that the role of `additional_hash` (https://github.com/raiden-network/raiden/issues/3082 )

This is a second shot after #135.

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -14,22 +14,15 @@ Data Structures
 Offchain Balance Proof
 ----------------------
 
-Data required by the smart contracts to update the payment channel end of the participant that signed the balance proof.
-Messages into smart contracts contain a shorter form called :ref:`Onchain Balance Proof <balance-proof-onchain>`.
+An offchain balance proof is a set of fields of Raiden protocol messages. The fields contain enough information for producing an :ref:`Onchain Balance Proof <balance-proof-onchain>`, which is required by the smart contracts to update the payment channel regarding the state of the participant that signed the balance proof. The offchain balance contains more information for easier bookkeeping.
 
-The signature must be valid and is defined as:
 
-::
-
-    ecdsa_recoverable(privkey, keccak256(balance_hash || nonce || additional_hash || channel_identifier || token_network_address || chain_id))
-
-Fields
-^^^^^^
-
+Fields of the Offchain Balance Proof message
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 +--------------------------+------------+--------------------------------------------------------------------------------+
 | Field Name               | Field Type |  Description                                                                   |
 +==========================+============+================================================================================+
-|  nonce                   | uint256    | Strictly monotonic value used to order transfers. The nonce starts at 1        |
+|  nonce                   | uint256    | Strictly monotonic value used to order balance proofs. The nonce starts at 1   |
 +--------------------------+------------+--------------------------------------------------------------------------------+
 |  transferred_amount      | uint256    | Total transferred amount in the history of the channel (monotonic value)       |
 +--------------------------+------------+--------------------------------------------------------------------------------+
@@ -41,13 +34,22 @@ Fields
 +--------------------------+------------+--------------------------------------------------------------------------------+
 |  channel_identifier      | uint256    | Channel identifier inside the TokenNetwork contract                            |
 +--------------------------+------------+--------------------------------------------------------------------------------+
-|  message_hash            | bytes32    | Hash of the message                                                            |
-+--------------------------+------------+--------------------------------------------------------------------------------+
-|  signature               | bytes      | Elliptic Curve 256k1 signature on the above data                               |
+|  signature               | bytes      | Elliptic Curve 256k1 signature described below                                 |
 +--------------------------+------------+--------------------------------------------------------------------------------+
 |  chain_id                | uint256    | Chain identifier as defined in EIP155                                          |
 +--------------------------+------------+--------------------------------------------------------------------------------+
 
+
+Signature of Balance Proofs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The signature field in the offchain balance proof is exactly the same thing as the signature in the onchain balance proof.  The signature must be valid and is defined as:
+
+::
+
+    ecdsa_recoverable(privkey, keccak256(balance_hash || nonce || additional_hash || channel_identifier || token_network_address || chain_id))
+
+In the signature, ``additional_hash`` witnesses the integrity of the whole Raiden message that contains the balance proof.
 
 Merkle Tree
 -----------
--- a/messaging.rst
+++ b/messaging.rst
@@ -49,7 +49,7 @@ The signature field in the offchain balance proof is exactly the same thing as t
 
     ecdsa_recoverable(privkey, keccak256(balance_hash || nonce || additional_hash || channel_identifier || token_network_address || chain_id))
 
-In the signature, ``additional_hash`` witnesses the integrity of the whole Raiden message that contains the balance proof.
+In the signature, :term:`additional_hash` witnesses the integrity of the whole Raiden message that contains the balance proof.
 
 Merkle Tree
 -----------
--- a/terminology.rst
+++ b/terminology.rst
@@ -172,7 +172,7 @@ Raiden Terminology
 
    Additional Hash
    additional_hash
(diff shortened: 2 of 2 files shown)


messaging.rst
```diff
@@ -14,22 +14,16 @@ Data Structures
...
-Fields
-^^^^^^
+Fields of the Offchain Balance Proof message
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 +--------------------------+------------+--------------------------------------------------------------------------------+
 | Field Name               | Field Type |  Description                                                                   |
 +==========================+============+================================================================================+
-|  nonce                   | uint256    | Strictly monotonic value used to order transfers. The nonce starts at 1        |
+|  nonce                   | uint256    | Strictly monotonic value used to order messages. The nonce starts at 1         |
```
pirapira: Now I'm not sure after looking into the code.  In the code, nonces are on balance proofs not on messages.

messaging.rst
```diff
@@ -14,22 +14,16 @@ Data Structures
 Offchain Balance Proof
 ----------------------
 
-Data required by the smart contracts to update the payment channel end of the participant that signed the balance proof.
-Messages into smart contracts contain a shorter form called :ref:`Onchain Balance Proof <balance-proof-onchain>`.
+An offchain balance proof is a set of fields of Raiden protocol messages.  The fields contain enough information for producing an :ref:`Onchain Balance Proof <balance-proof-onchain>`, which is required by the smart contracts to update the payment channel (especially, the state of the participant that signed the balance proof).  The offchain balance contains more information for easier bookkeeping.
```
LefterisJP: nitpick: 

> which is required by the smart contracts to update the payment channel (especially, the state of the participant that signed the balance proof)

As far as the contracts are concerned the state of the participant that signed the balance proof is the only thing that can be updated from this.

loredanacirstea: You have double space in two places: `messages.  The fields`, `proof).  The offchain`

messaging.rst
loredanacirstea: I agree that `transfers` is not optimal, as the `nonce` is increased multiple times during a `transfer`. I am fine with `messages` or even `balance proofs`.

messaging.rst
```diff
@@ -41,14 +35,23 @@ Fields
 +--------------------------+------------+--------------------------------------------------------------------------------+
 |  channel_identifier      | uint256    | Channel identifier inside the TokenNetwork contract                            |
 +--------------------------+------------+--------------------------------------------------------------------------------+
-|  message_hash            | bytes32    | Hash of the message                                                            |
-+--------------------------+------------+--------------------------------------------------------------------------------+
-|  signature               | bytes      | Elliptic Curve 256k1 signature on the above data                               |
```
loredanacirstea: Why was the `message_hash` removed from the fields? Also, it's worth keeping the `additional_hash` naming consistent. An issue can be opened in the `raiden` repo to make this consistent in the implementation (if not done already).

messaging.rst
```diff
@@ -41,14 +35,23 @@ Fields
...
+Signature of Balance Proofs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The signature field in the offchain balance proof is exactly the same thing as the signature in the onchain balance proof.  The signature must be valid and is defined as:
+
+::
+
+    ecdsa_recoverable(privkey, keccak256(balance_hash || nonce || additional_hash || channel_identifier || token_network_address || chain_id)
+
+In the signature, ``additional_hash`` witnesses the integrity of the whole Raiden message that contains the balance proof.
```
loredanacirstea: Consider adding a reference to the `additional_hash` terminology definition.

messaging.rst
pirapira: Will fix in this PR.

messaging.rst
pirapira: That's a good idea.  I will do this in this PR.

messaging.rst
pirapira: I'll go for `balance proofs`.

messaging.rst
pirapira: The implementation does not contain `message_hash` or `additional_hash` in the packed format.  For now, I'm following what @konradkonrad and @andrevmatos said in https://github.com/raiden-network/raiden/issues/3083 .  Once the discussion in https://github.com/raiden-network/raiden/issues/3083 settles, I can update this table.

messaging.rst
pirapira: Done!

pirapira:
Rebased and fixed conflicts.

messaging.rst
LefterisJP: @pirapira what exactly does this table want to document? From the name I assume the packed [balance_proof](https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/messages.py#L311) that is sent in the wire as part of all EnvelopeMessages?

If so perhaps you can also mention that the actual balance proof contains a `balance_hash` which is the hash of `transferred_amount` + `locked_amount` + `locksroot`. So they are not really part of the balance proof but hashed inside.

Also the offichain balance proof contains a `msg_type` which is intended to be consumed by the contract and contains one of [these](https://github.com/raiden-network/raiden-contracts/blob/c6e94549c7296204b61742da174ef102961bbeb3/raiden_contracts/contracts/TokenNetwork.sol#L105) values.

Finally the [additional_hash](https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/transfer/balance_proof.py#L30) is still there.

But perhaps you wanted to document some other message?

pirapira: @LefterisJP the result of `pack_balance_proof()` is not contained in any messages.  It's just for calculating hashes.

pirapira: 1. The spec says a `LockedTransfer` should contain an offchain balance proof.  https://github.com/raiden-network/spec/blob/4a2868c6728e7917d0847f68a0673135fdf10df0/messaging.rst#fields-3
2. The implementation says `LockedTransfer` does not contain an `additional_hash`.
https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/messages.py#L1102

These are why `additional_hash` is not part of an offchain balance proofl

LefterisJP: I meant it's contained as part of the signature in all the Balance proof containing messages since this is the data that is signed.

But still which message from [messages.py](https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/messages.py) is this table supposed to document then?

pirapira: This table documents `LockedTransfer - {lock, initiator, target}`.

LefterisJP: Hmm that does not seem right. I think what you are documenting is essentially the "Unlock" message, known in the codebase as [Secret](https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/messages.py#L645).

Cause if it was `LockedTransfer` you would also need a `token_address`, `lock_expiration`, `secret_hash`, `recipient` e.t.c. as seen [here](https://github.com/raiden-network/raiden/blob/dff9adfe0d9efe8d4dff46e457318cdbcc992125/raiden/messages.py#L1102)

Can you perhaps add a note that this is the message which you are trying to document for the implementation?

LefterisJP: > Maybe the whole thing is easier if we just document all fields of all real messages, without using this OffchainBalanceProof macro?

As discussed offline probably yes :)

pirapira:
>> Maybe the whole thing is easier if we just document all fields of all real messages, without using this OffchainBalanceProof macro?

> As discussed offline probably yes :)

I'll rework on this insight.

messaging.rst
pirapira: Since I closed this PR, I have to take care of what's discussed here.

pirapira: So I opened https://github.com/raiden-network/spec/pull/161

messaging.rst
loredanacirstea: > Maybe the whole thing is easier if we just document all fields of all real messages, without using this OffchainBalanceProof macro?

Agree. We had this format in the beginning and I am still in favor of it, especially that it triggered so many discussions on naming.

PR #137, opened 2018-11-28
Clarify spec of DirectTransfer message

pirapira:
I'm trying to clarify the specification of DirectTransfer messages:

* added some words in the validity condition, to relate the conditions with the surrounding environment.
* added two ID fields that I saw in the implementation but not in the spec.


LefterisJP:
@pirapira note: We are going to remove Direct Transfer from the implementation as it's actually not used at all. Once that is done perhaps we can also remove it from the spec. I got a branch in the client already working on this.

pirapira:
> going to remove Direct Transfer
I didn't know this.  Thanks.

pirapira:
I'll wait for https://github.com/raiden-network/raiden/pull/3125 and remove DirectTransfer from the spec.

PR #139, opened 2018-11-28
Change the description of additional_hash

pirapira:
I learned that additional_hash in balance proofs are used to protect the integrity of offchain messages.
https://github.com/raiden-network/raiden/issues/3082

So I'm updating the spec with this piece of knowledge.


smart_contracts.rst
```diff
@@ -554,7 +556,7 @@ Called after a channel has been closed. Can be called by any Ethereum address an
 - ``non_closing_participant``: Ethereum address of the channel participant who is updating the balance proof data.
 - ``balance_hash``: Hash of the balance data
 - ``nonce``: Strictly monotonic value used to order transfers.
-- ``additional_hash``: Computed from the message. Used for message authentication.
+- ``additional_hash``: Computed from the offchain message. Used for message authentication.
```
LefterisJP: Perhaps add also here the same note: That this hash can in the future also contain other application specific meta data?

pirapira: Done.

PR #140, opened 2018-11-28
Balance Proof Update not Balance Update Proof

pirapira:
Both "balance proof update"s and "balance update proof"s appeared in the specification.  I think it's a balance proof update because it's about updating the balance proof.

Also added a cross-reference.

### DIFF
--- a/smart_contracts.rst
+++ b/smart_contracts.rst
@@ -120,9 +120,9 @@ Balance Data Hash
 |  locksroot             | bytes32    | Root of merkle tree of all pending lock lockhashes                                    |
 +------------------------+------------+---------------------------------------------------------------------------------------+
 
-.. _balance-update-proof-message:
+.. _balance-proof-update-onchain:
 
-Balance Update Proof
+Balance Proof Update
 --------------------
 
 ::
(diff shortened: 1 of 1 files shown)


smart_contracts.rst
```diff
@@ -118,9 +118,9 @@ Balance Data Hash
 |  locksroot             | bytes32    | Root of merkle tree of all pending lock lockhashes                                    |
 +------------------------+------------+---------------------------------------------------------------------------------------+
 
-.. _balance-update-proof-message:
+.. _balance-proof-update-message:
```
loredanacirstea: Just a note: looking at this `balance-proof-update-message` reference now, after https://github.com/raiden-network/spec/pull/131, maybe the `message` part is not correct, because this does not describe the Raiden client message. Consider something like `balance-proof-update-onchain`.

smart_contracts.rst
```diff
@@ -566,7 +566,9 @@ Called after a channel has been closed. Can be called by any Ethereum address an
 
     Only a valid signed :term:`balance proof` from the channel's ``closing participant`` (the other channel participant) ``MUST`` be accepted. This :term:`balance proof` sets the amount of tokens owed to the ``non-closing participant`` by the ``closing participant``.
 
-    Only a valid signed balance proof update message  ``MUST`` be accepted. This message is a confirmation from the ``non-closing participant`` that the contained :term:`balance proof` can be set on his behalf.
+    Only a valid signed `balance proof update`__ message  ``MUST`` be accepted. This message is a confirmation from the ``non-closing participant`` that the contained :term:`balance proof` can be set on his behalf.
+
```
loredanacirstea: On the same note, the word `message` might be redundant here in the 2 places it occurs. Consider removing it.

smart_contracts.rst
pirapira: Will do in this PR.

smart_contracts.rst
pirapira: I see the word "message" in many similar sentences in this document.  I'll consider changing them together in a new issue https://github.com/raiden-network/spec/issues/144.

PR #169, opened 2018-12-12
Describe all fields of Unlock message

pirapira:
This commit
 - adds some fields that were missing before
 - adds a link to the corresponding part of code
 - adds an explanation what kind of signature is included

to the specification of the Unlock message.

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -187,14 +187,40 @@ Invariants
 Fields
 ^^^^^^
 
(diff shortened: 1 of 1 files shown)


messaging.rst
```diff
@@ -187,14 +187,40 @@ Invariants
 Fields
 ^^^^^^
 
+This should match `the Secret message in encoding/messages file <https://github.com/raiden-network/raiden/blob/a19a6c853b55f13725f2545c77b0475cbcc86807/raiden/encoding/messages.py#L113>`_.
```
LefterisJP: note: Once renamed we would need to change the text here to "Unlock"

pirapira: That falls into my task list (it's even shrinking these days!).

PR #171, opened 2018-12-13
Lefteris PFS-API feedback

karlb:
Changes in response to https://github.com/raiden-network/raiden-pathfinding-service/issues/83#issuecomment-447008346.

@LefterisJP @palango @Dominik1999: Are you all ok with this? If this gets merged, I'll update the PFS code accordingly.

### DIFF
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -186,8 +186,8 @@ Example
 ``api/1/<token_network_address>/paths``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The method will do ``num_paths`` iterations of Dijkstras algorithm on the last-known state of the Raiden
-Network (regarded as directed weighted graph) to return ``num_paths`` different paths for a mediated transfer of ``value``.
+The method will do ``max_paths`` iterations of Dijkstras algorithm on the last-known state of the Raiden
+Network (regarded as directed weighted graph) to return ``max_paths`` different paths for a mediated transfer of ``value``.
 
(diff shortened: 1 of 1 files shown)


pathfinding_service.rst
```diff
@@ -231,6 +231,10 @@ If no possible path is found, one of the following errors is returned:
 * No suitable path found
 * Rate limit exceeded
 * From or to invalid
+* The token network address is invalid
+* Bias is invalid
+* Max_path is invalid
```
palango: s/Max_path/max_paths

pathfinding_service.rst
```diff
@@ -267,11 +271,11 @@ Example
     }
     // Result for failure
     {
-        "error": "No suitable path found."
+        "errors": "No suitable path found."
```
palango: Should it be possible to have multiple errors here, e.g. `The token network address is invalid` and `Bias is invalid`, or just the first one?

pathfinding_service.rst
karlb: I wanted to stay in line with the previous errors. If we change that, then it should be
```
* No suitable path found
* Rate limit exceeded
* "from" or "to" invalid
* The "token_network_address" is invalid
* "bias" is invalid
* "max_path" is invalid
```

pathfinding_service.rst
karlb: From what I've seen, the raiden client only returns a single error. If we change that, we should do it consistently across all raiden repos.

LefterisJP: yes I think the idea in raiden was that we would in the future have multiple errors. So that this could either be a list or string.

LefterisJP:
Looks good from my side

PR #172, opened 2018-12-17
Update of the path finding spec according to current design

Dominik1999:
This is a draft of the path finding spec


pathfinding_service.rst
```diff
@@ -341,18 +177,11 @@ Example
...
-Network topology updates
+Network Topology Updates
 ------------------------
 
 .. Note::
-    A pathfinding service might want to cover multiple token networks. However,
-    it always needs to cover the *RDN* token network in order to be able to
-    provide routing information for payments.
-    
+    A pathfinding service might want to cover multiple token networks.
```
karlb: When do we have two separate PFSs? When they run in two processes? When they have to different ethereum addresses for payment? When they use two different IP addresses? So I'm not sure what this statement actually means.

pathfinding_service.rst
```diff
@@ -247,36 +113,6 @@ Example
         "value": 45,
         "max_paths": 10
     }'  /api/1/paths
-    // Request with specific preference
```
karlb: I like the curl examples and error output. Why don't we keep the parts that still apply?

pathfinding_service.rst
karlb: I guess my feedback was not very actionable in this case. I just wanted to state that the meaning of that sentence is unclear to me. We can either remove it or add more details to it.

Depending on our decisions regarding the payment, if might be helpful if a single ethereum address could be used to pay for routes in multiple token networks, since that could reduce the required number of payment channels if we go with a payment channel solution.

It's probably best to remove it until we know what exactly we want to say here, which I don't at this point.

Dominik1999: Thanks for the clarification @karlb !

pathfinding_service.rst
Dominik1999: reintegrated the curl examples and the errors output without fees

Issue #162, opened 2018-12-10
Add links from 'Message Flow' to terminologies


pirapira:
[The section "Message Flow" in Raiden Messages Specification](https://github.com/raiden-network/spec/blob/98eeadc41ad811adf5c02b541accdbb3b867d815/messaging.rst#mediated-transfer) contains names of messages (secret request, secret reveal etc.).  This issue keeps track of adding links from these names to the corresponding entries in the terminology section.

This issue spins off from
https://github.com/raiden-network/spec/pull/156#discussion_r240007709

PR #174, opened 2018-12-17
Refund Transfer messages in the terminology

pirapira:
This commit adds an entry for Refund Transfer messages in the terminology. Also a link to the entry is added.


pirapira:
This is a part of https://github.com/raiden-network/spec/issues/162.

terminology.rst
```diff
@@ -116,6 +116,10 @@ Raiden Terminology
    Locked Transfer message
        An offchain Raiden message that reserves an amount of tokens for a specific :term:`Payment`. See :ref:`locked-transfer-message` for details.
 
+   Refund Transfer
+   Refund Transfer message
+       An offchain Raiden message for a :term:`Payment` seeking a rerouting. When a receiver of a :term:`Locked Transfer` message gives up reaching the target, she returns a Refund Transfer message. A Refund Transfer message locks an amount of tokens in the same way as a :term:`Locked Transfer` message does.
```
LefterisJP: /s/she/they

LefterisJP: Also perhaps add "but in the opposite direction" in the end.

pirapira: I changed `she` into `they`.

I didn't add "the opposite direction" because, to me, "but in the opposite direction" sounded confusing.  (The sender of the Refund message can lock amounts coming from the receiver of the message?)

LefterisJP: What I meant is that for a refund transfer the receiver of the transfer sends a locked transfer with the same amounts back to the payer.

pirapira: I see.  That explains the name "refund" well.  Let me figure out something.

pirapira: Done.  I think this works: https://github.com/raiden-network/spec/pull/174/files#diff-a84d6f345909c58c851843f12aa76f5fR123.

pirapira:
@LefterisJP merge?

PR #182, opened 2019-01-02
Explain settle_block_number from getChannelInfo()

pirapira:
The meaning of ``settle_block_number`` is worth explaining because the number means different things according to different states of the channel.


pirapira:
I'll put more work on merging "settlement window" and "challenge period".

pirapira:
I thought "the closing participant has challenged the partner to show balance proofs".

PR #187, opened 2019-01-02
Add more hyperlinks to the terminology

pirapira:


terminology.rst
```diff
@@ -98,10 +100,10 @@ Raiden Terminology
        The node that receives a :term:`Payment`.
 
    Mediator
-       A node that mediates a transfer.
+       A node that mediates a :term:`Payment`.
```
loredanacirstea: I agree that `Payment` is correct here and not `transfer`.

PR #193, opened 2019-01-10
Remove REST Api from monitoring service

palango:
Following https://github.com/raiden-network/raiden-monitoring-service/pull/50


palango:
Included in #194

PR #200, opened 2019-01-23
Adds example values to the monitor request's spec

Dominik1999:


monitoring_service.rst
```diff
@@ -219,18 +219,18 @@ Example Data: Monitor Request
 
     {
       "onchain_balance_proof": {
-          "signature_prefix": tbd,
-          "message_length": 212,
-          "token_network_address": tbd,
-          "chain_id": tbd,
+          "signature_prefix": "13",
+          "message_length": "212",
```
palango: Is this a string?

palango: Seems like it has to be a string. Can you make this clearer in the **Note** box:

Length of the actual message to be signed in decimal representation (not null-terminated).

to

String of the length of the actual message to be signed in decimal representation (not null-terminated).

Issue #120, opened 2018-09-03
Overview of Raiden Network RSB (non-core) smart contracts


loredanacirstea:
Notes from the 24 August planning meeting regarding this:

`MSC` - Monitoring Service Contract - https://github.com/raiden-network/spec/issues/46
`UDC` - User Deposit Contract
`PFSC` - PathFinding Service Contract
`RSBC` - RaidenServiceBundle Contract
`HC` - Hub Contract

## User Deposit Contract
- global deposit in RDN for all services
- `MSC` will transfer the user’s tokens from here, when it has to pay a reward for the MS
- this will mark the `MSC`, `PFSC`, `RSBC` as trusted contracts
- if the user wants to withdraw part/all of his deposit, this must be timed out (e.g. he can apply for a withdrawal and then finalized it only after 1 week); the reasoning is: the services must make sure they withdraw their payments first. `PFS` will specifically be targeted here, if we implement off-chain payments.
- the contract will make sure the 3rd parties withdrawing from it are registered in `RSBC`
- PFS, MS will be paid from this contract

## RaidenServiceBundle Contract
- global deposit in RDN for 3rd party services (MS, PFS, Matrix HS); each 3rd party will deposit/bound tokens here.
- no slashing was discussed

## PFSC
- it can act as a unidirectional payment channel between the users and PFS, using the user's global deposit from `UDC`

## Hub Contract

Ideally, the user must have enough RDN to use the system. The onboarding process must use this RDN to set the user up.

The onboarding transactions are:
- RDN deposit to `UDC`
- token of choice channel with a Hub

This is why we introduce a `Hub Contract`, that will implement 1 onboarding transaction in the Raiden Network between User -> HC. This transaction will:
- approve deposit token amount to be sent to `TokenNetwork`; token will be whatever the user chooses, as long as the Hub also supports it
- create User - Hub channel (`TokenNetwork`) (! Hub currently needs an Ethereum address with which to sign messages)
- User deposit (`TokenNetwork`)
- Hub’s deposit (`TokenNetwork`)

Note: the `HC` can be the one that sends the tokens for both his & the user's deposit. He can take `RDN` out from the `UDC` for the user deposit + some fee that he requests. So, he also does some exchange calculations (`RDN` - token_of_choice).

hackaugusto:
> User Deposit Contract

pros:
1. This will *require* only one transaction from *the user* to pay for the third party services (the funding transaction, additional transactions may be done to top up or withdraw, but are not *required*)
[...]

PR #202, opened 2019-01-25
Initial service contract docs + autosol

karlb:
Goals of this commit:
* Provide basis for more detailed service contract docs
* Incorporate https://github.com/raiden-network/spec/issues/120
* Incorporate https://github.com/raiden-network/raiden-contracts/issues/447
* Allow automatic generation of docs from solidity docstrings

More details, links, diagrams, etc. will follow in the next PR.


Makefile
```diff
@@ -47,47 +47,52 @@ help:
 
 .PHONY: clean
 clean:
-	rm -rf $(BUILDDIR)/*
+	rm -rf $(BUILDDIR)/* contracts
+
+# download `raiden-contracts/raiden_contracts/contracts` to the local `contracts` dir
+# This is skipped if the dir already exists. Use `make clean` to force an update.
+contracts:
+	curl -sSL https://github.com/raiden-network/raiden-contracts/tarball/master | tar xvz --wildcards '*/raiden_contracts/contracts' --strip-components 2
```
palango: I get the following error on macOS, can we workaround that?
```sh
curl -sSL https://github.com/raiden-network/raiden-contracts/tarball/master | tar xvz --wildcards '*/raiden_contracts/contracts' --strip-components 2
tar: Option --wildcards is not supported
```

palango: Additionally, is this run on readthedocs?

karlb: Apparently `--wildcards` is a gnu extension and the BSD tar is missing that. I'll find an alternative way of doing that.

Why should `readthedocs` be a problem? I thought it was just a linux system with sphinx installed that runs the different `make` targets.

karlb: Please try again on MacOS, now.

PR #205, opened 2019-01-29
Add OneToN spec

karlb:
This is https://github.com/raiden-network/raiden-services/issues/38 with
only few changes. Spefic functions will be documented as the
implemenation makes progress.

### DIFF
--- a/service_contracts.rst
+++ b/service_contracts.rst
@@ -63,7 +63,148 @@ The PFS will be paid with signed IOUs, roughly a simplified uRaiden adapted to 1
 OneToN
 ======
 
-Specification in progress. See https://github.com/raiden-network/raiden-services/issues/38.
+Overview
+--------
+
+The OneToN contract handles payments for the PFS. It has been chosen with the
+following properties in mind:
+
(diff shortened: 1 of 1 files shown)


service_contracts.rst
```diff
@@ -63,7 +63,148 @@ The PFS will be paid with signed IOUs, roughly a simplified uRaiden adapted to 1
...
+-  reliability, high probability of success (P > 0.99)
+-  low cost overhead (<5% of transferred amount)
+-  low fraud rate (< 3%, i.e. some fraud is tolerable)
+-  can be implemented quickly
+
+Communication between client and PFS
+------------------------------------
+
+When requesting a route, the IOU is added as five separate arguments to
+the `existing HTTP query params`_.
```
palango: I though it would go in the request header as it's done in uRaiden!? Does this bring advantages?

karlb: And in this case GET is not a good idea, since we expect different results and different payment info each time.

palango: Ok, so we need to look into that anyways.

PR #206, opened 2019-01-29
Add Isabelle proof script on the settlement algo

pirapira:
This is how I checked claims in the spec about the settlement algorithms.

### DIFF
--- a/proofs/Settlement.thy
+++ b/proofs/Settlement.thy
@@ -1,0 +1,160 @@
+theory Settlement
+
+imports Main
+
+begin
+
+text "This file contains an analysis of the settlement algorithm in the TokenNetwork contract.
+https://raiden-network-specification.readthedocs.io/en/latest/smart_contracts.html#protocol-values-and-settlement-algorithm-analysis
+The result so far confirms two things:
+  lemma s1_correct: The value calculated as 'S1 = RmaxP1 - SL2' is equal to 'S1 = D1 - W1 + T2 - T1 - L1'.
+  lemma s2_correct: Similarly for 'S2 = RmaxP2 - SL1' and 'S2 = D2 - W2 + T1 - T2 - L2'.
+The required conditions appear in the statements of the lemma.
+"
+
+text "TODO:
+* Make sure that, you can only lose tokens if you submit an older balance proof.
+"
+
+type_synonym impl_number = "int option"
+
+definition valid :: "int \<Rightarrow> bool"
+  where "valid a = (0 \<le> a \<and> a < 32)"
+
+definition chop :: "int \<Rightarrow> impl_number"
+  where
+    "chop a = (if valid a then Some a else None)" (* has to change it to min/max*)
+
+fun impl_add :: "impl_number \<Rightarrow> impl_number \<Rightarrow> impl_number"
+  where
+  "impl_add None _ = None"
+| "impl_add _ None = None"
+| "impl_add (Some a) (Some b) =
+     chop (a + b)"
+
+value "impl_add (Some 10) (Some 25)"
+value "impl_add (Some 10) (Some 1)"
+
+
+fun impl_sub :: "impl_number \<Rightarrow> impl_number \<Rightarrow> impl_number"
+  where
+  "impl_sub None _ = None"
+| "impl_sub _ None = None"
+| "impl_sub (Some a) (Some b) =
+    chop (a - b)"
+
+value "impl_sub (Some 100) (Some 200)"
+
+
+fun impl_min :: "impl_number \<Rightarrow> impl_number \<Rightarrow> impl_number"
+  where
+  "impl_min None _ = None"
+| "impl_min _ None = None"
+| "impl_min (Some a) (Some b) =
+    chop (min a b)"
+
+value "impl_sub (Some 100) (Some 200)"
+
+
+(*** settlement algorithm ***)
+
+definition TLmax1 :: "int \<Rightarrow> int \<Rightarrow> impl_number" where
+"TLmax1 T1 L1 = impl_add (Some T1) (Some L1)"
+
+definition TLmax2 :: "int \<Rightarrow> int \<Rightarrow> impl_number" where
+"TLmax2 T2 L2 = impl_add (Some T2) (Some L2)"
+
+definition RmaxP1_pre :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"RmaxP1_pre T1 L1 T2 L2 D1 W1 =
+  impl_sub (impl_add (impl_sub (TLmax2 T2 L2) (TLmax1 T1 L1)) (Some D1)) (Some W1)"
+
+definition TAD :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"TAD D1 D2 W1 W2 = impl_sub (impl_sub (impl_add (Some D1) (Some D2)) (Some W1)) (Some W2)"
+
+definition RmaxP1 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"RmaxP1 T1 L1 T2 L2 D1 W1 D2 W2 =
+  impl_min (TAD D1 D2 W1 W2) (RmaxP1_pre T1 L1 T2 L2 D1 W1)"
+
+definition SL2 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"SL2 T1 L1 T2 L2 D1 W1 D2 W2 =
+   impl_min (RmaxP1 T1 L1 T2 L2 D1 W1 D2 W2) (Some L2)"
+
+definition S1 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"S1 T1 L1 T2 L2 D1 W1 D2 W2 =
+   impl_sub (RmaxP1 T1 L1 T2 L2 D1 W1 D2 W2) (SL2 T1 L1 T2 L2 D1 W1 D2 W2)"
+
+definition RmaxP2 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number"
+  where
+"RmaxP2 T1 L1 T2 L2 D1 W1 D2 W2 =
+    impl_sub (TAD D1 D2 W1 W2) (RmaxP1 T1 L1 T2 L2 D1 W1 D2 W2)"
+
+definition SL1 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number"
+  where
+"SL1 T1 L1 T2 L2 D1 W1 D2 W2 = impl_min (RmaxP2 T1 L1 T2 L2 D1 W1 D2 W2) (Some L1)"
+
+definition S2 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number"
+  where
+"S2 T1 L1 T2 L2 D1 W1 D2 W2
+ = impl_sub (RmaxP2 T1 L1 T2 L2 D1 W1 D2 W2) (SL1 T1 L1 T2 L2 D1 W1 D2 W2)"
+
+definition spec_s1 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int"
+  where
+"spec_s1 T1 T2 D1 W1 L1 = D1 - W1 + T2 - T1 - L1"
+
+definition spec_s2 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int"
+  where
+"spec_s2 T1 T2 D2 W2 L2 = D2 - W2 + T1 - T2 - L2"
+
+lemma s1_correct :
+"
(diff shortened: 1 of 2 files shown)


proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+"TLmax2 T2 L2 = impl_add (Some T2) (Some L2)"
+
+definition RmaxP1_pre :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"RmaxP1_pre T1 L1 T2 L2 D1 W1 =
+  impl_sub (impl_add (impl_sub (TLmax2 T2 L2) (TLmax1 T1 L1)) (Some D1)) (Some W1)"
+
+definition TAD :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
+"TAD D1 D2 W1 W2 = impl_sub (impl_sub (impl_add (Some D1) (Some D2)) (Some W1)) (Some W2)"
+
+definition RmaxP1 :: "int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> int \<Rightarrow> impl_number" where
```
pirapira: These lines look terrible as ASCII texts.  The Isabelle IDE automatically shows this as
```
definition RmaxP2 :: "int ⇒ int ⇒ int ⇒ int ⇒ int ⇒ int ⇒ int ⇒ int ⇒ impl_number"
```

I know it's a peculiar engineering choice.

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+valid D1 \<Longrightarrow>
+valid D2 \<Longrightarrow>
+valid (D1 + D2) \<Longrightarrow> (* (12) *)
+valid W1 \<Longrightarrow>
+valid W2 \<Longrightarrow>
+valid (T1 + L1) \<Longrightarrow> (* 10 *)
+valid (T2 + L2) \<Longrightarrow> (* 10 *)
+L1 <= D1 - W1 + T2 - T1 \<Longrightarrow> (* (5) *)
+D1 - W1 + T2 - T1 - L1 \<le> D1 + D2 - W1 - W2 \<Longrightarrow> (* (5) *)
+-(D1 - W1) <= T2 + L2 - T1 - L1 \<Longrightarrow> (* (7) *)
```
hackaugusto: Hummm, the rule `( 7 )` from the spec does not seem to sustain itself on its own, I think it should be broken down into two.

The current statement is `(7 R) -(D1 - W1) <= T2 + L2 - T1 - L1 <= D2 - W2`, which boils down to two things (note that there is only one locked amount):

- A node cannot commit/send more then it has `-(D1 - W1) <= T2 - T1 - L1`
- A node must not accept from its partner more then he has `T2 + L2 - T1 <= D2 - W2`

Having both locked amounts in there would make this valid:

`W1=0`, `W2=0`, `T1=0`, `T2=0`, reduces `-(D1 - W1) <= T2 + L2 - T1 - L1 <= D2 - W2` to `-D1 <= L2 - L1 <= D2`

D1 | D2 | L1 | L2 | -D1 <= L2 - L1 <= D2
--- | --- | --- | --- | ---
1 | 1 | 0 | 0 | true
1 | 1 | 1 | 0 | true
1 | 1 | 2 | 1 | true

The last line above should not be true, however, with `-(D1 - W1) <= T2 - T1 - L1`

D1 | L1| -D1 <= - L1
--- | --- | ---
1 | 0 | true
1 | 1 | true
1 | 2 | false

I believe the above worked because of the stricter definition just above `L1 <= D1 - W1 + T2 - T1` (but I have not checked this hypothesis)

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+"
+
+text "TODO:
+* Make sure that, you can only lose tokens if you submit an older balance proof.
+"
+
+type_synonym impl_number = "int option"
+
+definition valid :: "int \<Rightarrow> bool"
+  where "valid a = (0 \<le> a \<and> a < 32)"
```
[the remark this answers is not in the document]
pirapira: You're right.

It's `[0, 32)` only because Isabelle finds counter-examples faster with the smaller region (but that's relevant only when the proof fails).

I'll try to change this into `[0, 2**256)` and see if the proofs still go through.

pirapira:
> Q: Does the order of expressions matter for Isabelle? This is a big point of concern for the solidity code, as each operation may potentially overflow/underflow. For instance, TLmax2 - TLmax1 + D1 - W1, has to be computed from left to right, otherwise overflow/underflow are possible (e.g. TLmax2 + D1 will overflow). IOW, isn't necessary to have valid expressions for each intermediary value?

No, `a - b + c` is always equal to `c + a - b` and so on.

So I defined `impl_add` and `impl_sub` that overflow and underflow.  `S1` and `S2` use these overflowing&underflowing operations.

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+valid T2 \<Longrightarrow>
+valid L1 \<Longrightarrow>
+valid L2 \<Longrightarrow>
+valid D1 \<Longrightarrow>
+valid D2 \<Longrightarrow>
+valid (D1 + D2) \<Longrightarrow> (* (12) *)
+valid W1 \<Longrightarrow>
+valid W2 \<Longrightarrow>
+valid (T1 + L1) \<Longrightarrow> (* 10 *)
+valid (T2 + L2) \<Longrightarrow> (* 10 *)
```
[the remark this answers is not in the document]
pirapira: You're right.  I'll make a PR to the spec.

proofs/Settlement.thy
pirapira: Now I tried if Isabelle understands `0 <= 1 <= 2`; no it doesn't.

I guess
```
 -(D1 - W1) <= T2 + L2 - T1 - L1 /\ T2 + L2 - T1 - L1 <= D2 - W2 (7 R)
```
is the best approximation.  I'll use ^ unless I hear otherwise.

proofs/Settlement.thy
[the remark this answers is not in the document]
loredanacirstea: There is:
> (11 R) T1 + L1 < 2^256 ; T2 + L2 < 2^256

What I don't have is the individual constraints - e.g. `T1 < 2^256`. I did not add them because this is handled by the EVM, but for the sake of completeness, we can add them. Suggestion: before `Protocol Values Constraints`, as `Implementation/Solidity-enforced Values Constraints` or similar.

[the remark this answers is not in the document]
pirapira: I see I made the numbering mistake.  Will fix.

For individual `T1`, `T2`, etc I'll add comments `because of EVM & Solidity types` in the Isabelle script.

proofs/Settlement.thy
loredanacirstea: The 3rd case seems weird:

D1 | D2 | L1 | L2 | -D1 <= L2 - L1 <= D2
--- | --- | --- | --- | ---
1 | 1 | 2 | 1 | true

because to get to that case, we need a state of:

D1 | D2 | L1 | L2
--- | --- | --- | --- 
1 | 1 | 1 | 1

Here, both available balances are `0`:  `AB1 = 1 - 0 + 0 - 0 - 1 = 0` & `AB2 = 1 - 0 + 0 - 0 - 1 = 0`. So, none of the participants can make transfers -> under `(5 R)` you can't have `L1 == 2`.
Am I missing something?

loredanacirstea:
I assume the order is actually enforced by doing `impl_sub (impl_add (impl_sub (TLmax2 T2 L2) (TLmax1 T1 L1)) (Some D1)) (Some W1)` and from what I see, you closely followed the Solidity implementation.

The order can be important if our Solidity implementation detects an underflow while the Isabelle proof does not, because it uses another order that neutralizes it. But I don't think it's the case here.

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+L1 <= D1 - W1 + T2 - T1 \<Longrightarrow> (* (5) *)
+D1 - W1 + T2 - T1 - L1 \<le> D1 + D2 - W1 - W2 \<Longrightarrow> (* (5) *)
+-(D1 - W1) <= T2 + L2 - T1 - L1 \<Longrightarrow> (* (7) *)
+T2 + L2 - T1 - L1 <= D2 - W2 \<Longrightarrow> (* (8) *)
```
loredanacirstea: Also `(* (7) *)`

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+imports Main
+
+begin
+
+text "This file contains an analysis of the settlement algorithm in the TokenNetwork contract.
+https://raiden-network-specification.readthedocs.io/en/latest/smart_contracts.html#protocol-values-and-settlement-algorithm-analysis
+The result so far confirms two things:
+  lemma s1_correct: The value calculated as 'S1 = RmaxP1 - SL2' is equal to 'S1 = D1 - W1 + T2 - T1 - L1'.
+  lemma s2_correct: Similarly for 'S2 = RmaxP2 - SL1' and 'S2 = D2 - W2 + T1 - T2 - L2'.
+The required conditions appear in the statements of the lemma.
```
loredanacirstea: With the notes that I already made in person:
- `s1_correct` & `s2_correct` - stand only if both balance proofs are `valid` & `last`; indeed, it should be always correct in this case.
- for cases where **at least** one balance proof is `old`, we need to move the check down the channel lifecycle: after both `unlock`s are done. In these cases, after settlement you can frequently have a different `S1` or `S2` value in the contract than `S1 = D1 - W1 + T2 - T1 - L1`, due to under/overflows. We need to make sure this difference balances out with the `unlock`s. 
  - Python tests reference: https://github.com/raiden-network/raiden-contracts/blob/76758d271059c96a8781754df3c599c4c948a2de/raiden_contracts/tests/test_channel_settle_unlock_state.py#L232-L244

pirapira: I have an issue about extending the Isabelle proof for older balance proofs https://github.com/raiden-network/spec/issues/207 .

pirapira:
@loredanacirstea 

> I **assume** the order is actually enforced by doing impl_sub (impl_add (impl_sub (TLmax2 T2 L2) (TLmax1 T1 L1)) (Some D1)) (Some W1)

Your assumption is correct.

proofs/Settlement.thy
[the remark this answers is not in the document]
pirapira: When I replace 32 with 2**256, the proof about S1 goes through but S2 fails.  Investigating...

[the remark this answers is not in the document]
pirapira: Now I need `(D1 + D2) < 2 ** 256`.  Not sure why.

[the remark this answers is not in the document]
pirapira: Done changing into "2 ** 256*.

> question, why do you use 0 \<le> a on the right, but a < 32 on the left?

Because 0 is a `uint256` but `2 ^ 256` is no longer a `uint256`.

proofs/Settlement.thy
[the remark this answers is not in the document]
pirapira: Now I have fixed the number `valid (T1 + L1) \<Longrightarrow> (* (11 R) *)`.

Also I've added comments to individual variables: `because it's uint256`.

proofs/Settlement.thy
```diff
@@ -0,0 +1,160 @@
...
+valid L1 \<Longrightarrow>
+valid L2 \<Longrightarrow>
+valid D1 \<Longrightarrow>
+valid D2 \<Longrightarrow>
+valid (D1 + D2) \<Longrightarrow> (* (12) *)
+valid W1 \<Longrightarrow>
+valid W2 \<Longrightarrow>
+valid (T1 + L1) \<Longrightarrow> (* 10 *)
+valid (T2 + L2) \<Longrightarrow> (* 10 *)
+L1 <= D1 - W1 + T2 - T1 \<Longrightarrow> (* (5) *)
```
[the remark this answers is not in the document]
pirapira: Now I have something closer to the spec.
```
D1 - W1 + T2 - T1 - L1 \<ge> 0 \<Longrightarrow>  (* (5 R) *)
D1 - W1 + T2 - T1 - L1 \<le> D1 + D2 - W1 - W2 \<Longrightarrow> (* (5 R) *)
```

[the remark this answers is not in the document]
pirapira: I checked the forms of the other conditions too.

proofs/Settlement.thy
pirapira: If this discussion is about changing the condition in the spec, please take it there.

proofs/Settlement.thy
pirapira: Fixed.

pirapira:
@hackaugusto @loredanacirstea this PR is ready for review again.

proofs/Settlement.thy
```diff
@@ -0,0 +1,155 @@
...
+valid (T1 + L1) \<Longrightarrow> (* (11 R) *)
+valid (T2 + L2) \<Longrightarrow> (* (11 R) *)
+valid (D1 + D2) \<Longrightarrow> (* (12) *)
+D1 - W1 + T2 - T1 - L1 \<ge> 0 \<Longrightarrow>  (* (5 R) *)
+D1 - W1 + T2 - T1 - L1 \<le> D1 + D2 - W1 - W2 \<Longrightarrow> (* (5 R) *)
+-(D1 - W1) <= T2 + L2 - T1 - L1 \<and> T2 + L2 - T1 - L1 <= D2 - W2 \<Longrightarrow> (* (7 R) *)
+T2 + L2 \<ge> T1 + L1 \<Longrightarrow>
+S1 T1 L1 T2 L2 D1 W1 D2 W2 = (Some (spec_s1 T1 T2 D1 W1 L1))"
+  by(auto simp add: valid_def spec_s1_def S1_def RmaxP1_def RmaxP1_pre_def TAD_def chop_def
+ TLmax2_def SL2_def TLmax1_def chop_def )
```
loredanacirstea: nitpick: duplicate `chop_def` (trying it out in Isabelle)

proofs/Settlement.thy
loredanacirstea: updated my comment - I missed the point first time I read it.
So, just to conclude:
So `-(D1 - W1) <= T2 + L2 - T1 - L1 <= D2 - W2` is useful in the demonstration for `T2 + L2 - T1 - L1 <= D2 - W2` - this is used solve the `+ D1` overflow from `(T2 + L2) - (T1 + L1) + D1 - W1`.
Agreed that the first part `-(D1 - W1) <= T2 + L2 - T1 - L1` is less strict than needed to sustain itself. This was added after this discussion: https://github.com/raiden-network/spec/pull/119#discussion_r214018326.
They can be separated if needed.

pirapira: Yes, I dropped the first part and the proof still works.

pirapira: And I opened an issue about changing the spec https://github.com/raiden-network/spec/issues/209 .

pirapira:
@loredanacirstea has the Isabelle IDE on her machine, so can also edit the script.

PR #212, opened 2019-02-04
Rename RSB in diagram, too

karlb:
Forgot this in the last PR.


palango:
Is the image generated automatically?

karlb:
Unfortunately, no. There's https://www.npmjs.com/package/nomnoml-cli, but I didn't want to touch npm just for this. If only used nomnoml for consistency with the existing docs. There are sphinx plugins for some diagram types that provide proper integration, but nomnoml is not among them.

PR #213, opened 2019-02-04
Simplify monitor request

hackaugusto:
Changes:

- A monitor service job is to watch an open channel, therefore there are
*no* on-chain balance proofs available yet. Because of this the field
`onchain_balance_proof` was renamed to simply `balance_proof`.
- A single signature is sufficient for the message to be authenticated, a
second signature is not only unecessary but also wasteful.


hackaugusto:
What is the plan for compatibility with the current version of the smart contracts? Because this change will most likely required changes to the smart contract to support the monitoring service request.

hackaugusto:
From the diff, I'm removing `reward_proof_signature`. The signature used to call `updateNonClosingBalanceProof` and for the reward proof can be the same.

monitoring_service.rst
```diff
@@ -196,7 +196,7 @@ Monitor Requests are generated by clients from `Onchain Balance Proofs`__.
 +--------------------------+------------+--------------------------------------------------------------------------------+
 | Field Name               | Field Type |  Description                                                                   |
 +==========================+============+================================================================================+
-|  onchain_balance_proof   | object     | Onchain Balance Proof                                                          |
+|  balance_proof           | object     | Latest Blinded Balance Proof  to be used by the monitor service                |
```
palango: double space after *Proof*

hackaugusto: fixed

PR #220, opened 2019-02-21
IOU endpoint for PFS

karlb:
TODO:

- [x] provide a way to the client to get the latest IOU

Postponed to later PRs:
- add error codes and add them to the examples

### DIFF
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -65,6 +65,8 @@ Network (regarded as directed weighted graph) to return ``max_paths`` different 
 
 * Applies on the fly changes to the graph's weights - depends on ``DIVERSITY_PEN_DEFAULT`` from ``config``, to penalize edges which are part of a path that is returned already.
 
+.. _path_args:
+
 Routing Arguments
 """""""""""""""""
 
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -85,7 +85,7 @@ The arguments are POSTed as a JSON object.
 +----------------------+---------------+-----------------------------------------------------------------------+
 | max_paths            | int           | The maximum number of paths returned.                                 |
 +----------------------+---------------+-----------------------------------------------------------------------+
-| fee_iou              | object        | IOU objects as described in :ref:`pfs_payment` to pay the service fee |
+| fee_iou              | object        | IOU object as described in :ref:`pfs_payment` to pay the service fee  |
 +----------------------+---------------+-----------------------------------------------------------------------+
 
 Returns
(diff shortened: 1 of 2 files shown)


pathfinding_service.rst
```diff
@@ -219,6 +223,47 @@ Example
...
+                                             || sender || receiver || timestamp ))
+
+Returns
+"""""""
+A JSON object with a single property:
+
++----------------------+---------------+-----------------------------------------------+
+| Field Name           | Field Type    | Description                                   |
++======================+===============+===============================================+
+| last_iou             | object        | IOU object as described in :ref:`pfs_payment` |
```
rakanalh: So this would be:
```
{
   "last_iou": {
        "fields1": ...,
        "fields2": ...
    }
}
```
Wouldn't be easier to just return the IOU object itself without nesting?

karlb: Yes, but this way we can easily add new fields like `last_usage_time` or `unclaimed_ious` without breaking compatibility or adding new endpoints.

pathfinding_service.rst
```diff
@@ -219,6 +223,47 @@ Example
...
+"""""""""
+
++---------------------+------------+---------------------------------------------------------+
+| Field Name          | Field Type | Description                                             |
++=====================+============+=========================================================+
+| sender              | address    | Sender of the payment (Ethereum address of client)      |
++---------------------+------------+---------------------------------------------------------+
+| receiver            | address    | Receiver of the payment (Ethereum address of PFS)       |
++---------------------+------------+---------------------------------------------------------+
+| timestamp           | string     | Current UTC date and time in ISO 8601 format            |
```
palango: Why is the timestamp needed?

pathfinding_service.rst
rakanalh: I see... good point.

pathfinding_service.rst
karlb: Without it anyone who somehow got hold of a request could replay it indefinitely to get information about IOUs for this user. This way, the PFS can check that the request is recent and refuse old requests.

palango: Ok, makes sense.

palango: Should we propose a maximum time delta after which the request should be ignored. Something like 10s?

karlb: Instructing the client to send the current time is enough for the spec IMO. If we want to specify something, I suggest to be less strict (e.g. 10min) since this only guards against long term information leaks. I don't want to prevent clients from working only because their local clock is off by a minute.

karlb:
I'm still missing an approval for this PR. Is there anything I have to do?

Travis failed for some internal reason and I don't have sufficient permissions to restart. What should I do in such a case?
* Nothing
* Force push without meaningful changes to trigger the build again
* Request admin permissions

rakanalh:
Looks good on my side... merge at your convenience

PR #223, opened 2019-03-15
Fix table markup

palango:


palango:
Yes, this work locally but it seems like github doesn't show it.

PR #225, opened 2019-03-25
Changes "fee" inot "reward" in the ms spec

Dominik1999:


monitoring_service.rst
```diff
@@ -4,11 +4,11 @@ Raiden Monitoring Service
 
 Summary
 =======
-*   Monitoring Service (MS) listens to blinded Balance Proofs (BP) and proposed service fees in a public Matrix Room
-*   Any pre-registered MS can decide to monitor a channel and store corresponding BPs
+*   Monitoring Service (MS) listens to Monitor Requests (MRs - blinded Balance Proofs) and proposed rewards in a public Matrix Room
```
palango: Maybe *offered* instead of *proposed*. *proposed* for me implies a possibility for negotiation, which isn't available.

PR #228, opened 2019-04-11
Update the PFS documentation

karlb:
...which was quite outdated in some parts.

* Fix totally wrong/outdated description of the info endpoint
* Bring name of `iou` parameter in line with current implementation
* Update routing preference description

### DIFF
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -29,16 +29,15 @@ High-Level-Description
 A node can request a list of possible paths from start point to endpoint for a given transfer value.
 The ``get_paths`` method implements the canonical Dijkstra algorithm to return a given number of paths
 for a mediated transfer of a given value. The design regards the Raiden network as an unidirectional
-weighted graph, where the default weights and therefore the primary constraint of the optimization)
-* at step (1 and 2) are 1 (no fees being implemented) and
-* at step (3) are the fees of each channel.
+weighted graph, where the weights of the edges/channels are the sum of multiple penalty terms:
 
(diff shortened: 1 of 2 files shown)


Dominik1999:
Thanks for the Update Karl!

karlb:
> Could you include the "canonical identifier" into the Balance Update. Within the canonical identifier, there are the fields "chain_id", "token_network_identifier" and "channel_identifier"

I didn't deal with the canonical identifier, yet. I think it's better if you do it in a separate PR.

pathfinding_service.rst
```diff
@@ -85,7 +84,13 @@ The arguments are POSTed as a JSON object.
 +----------------------+---------------+-----------------------------------------------------------------------+
 | max_paths            | int           | The maximum number of paths returned.                                 |
 +----------------------+---------------+-----------------------------------------------------------------------+
-| fee_iou              | object        | IOU object as described in :ref:`pfs_payment` to pay the service fee  |
+| iou                  | object        | IOU object as described in :ref:`pfs_payment` to pay the service fee  |
++----------------------+---------------+-----------------------------------------------------------------------+
+| diversity_penalty    | float         | (optional) Routing penalty per channel that is reused across multiple |
```
palango: Should we say something about the defaults here?

karlb: I wouldn't until we have some real world experience in tuning those values, so that we don't have to update the specs every time we change a default value. Also we might want to vary the defaults within a certain range once we're able to gather routing feedback.

PR #232, opened 2019-05-28
Add spec for routing feedback.

palango:
Closes https://github.com/raiden-network/raiden-services/issues/336

### DIFF
--- a/pathfinding_service.rst
+++ b/pathfinding_service.rst
@@ -39,18 +39,11 @@ weighted graph, where the weights of the edges/channels are the sum of multiple 
 
 See `Routing Preferences`_ for information on how to configure the trade-off between these penalties.
 
-Public Interfaces
-=================
-The pathfinding service provides four public interfaces
 
-* a public endpoint for path requests by Raiden nodes
-* an endpoint to get updates from blockchain events
-* an endpoint to send ``Capacity Updates`` to
+Public REST API
+===============
 
-Public Endpoints
-----------------
-
-A path finding service must provide the following endpoints. The interface has to be versioned.
+A pathfinding service must provide the following endpoints. The interface has to be versioned.
 
 The examples provided for each of the endpoints is for communication with a REST endpoint.
 
@@ -108,7 +101,8 @@ Each path object consists of the following information:
 | estimated_fee        | int           | An estimate of the fees required for that path.                       |
 +----------------------+---------------+-----------------------------------------------------------------------+
 
-The feedback token is the 32-character hexadecimal string representation of a UUID.
+The feedback token is the 32-character hexadecimal string representation of a UUID and is valid for all routes that
+are included in the response.
 
 Routing Preferences
 """""""""""""""""""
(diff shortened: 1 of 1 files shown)


pathfinding_service.rst
```diff
@@ -41,11 +41,11 @@ See `Routing Preferences`_ for information on how to configure the trade-off bet
...
 =================
-The path finding service needs three public interfaces
+The pathfinding service provides four public interfaces
 
-* a public endpoint for path requests by Raiden nodes
-* an endpoint to get updates from blockchain events
-* an endpoint to get updates about current balances and fees
+* A public REST endpoint for requests by Raiden nodes
+* An interface to get updates from the blockchain
+* An endpoint to receive broadcasted updates about channel balances and mediation fees
```
Dominik1999: here you list just three public interfaces, right?

pathfinding_service.rst
```diff
@@ -41,11 +41,11 @@ See `Routing Preferences`_ for information on how to configure the trade-off bet
...
 
 Public Endpoints
 ----------------
```
Dominik1999: here we say public endpoint whereas we say public interfaces above

pathfinding_service.rst
```diff
@@ -95,7 +95,10 @@ The arguments are POSTed as a JSON object.
 
 Returns
 """""""
-A list of path objects. A path object consists of the following information:
+
+A list of path objects and a feedback token when successful.
+
```
Dominik1999: just to be nitpicky but who knows what a path object is? Maybe we can find a more common term or describe that one?

pathfinding_service.rst
```diff
@@ -95,7 +95,10 @@ The arguments are POSTed as a JSON object.
...
+Each path object consists of the following information:
```
Dominik1999: is the content of the table what the path object consists of? So each path object is a list of tuples of a path and the corresponding fee. This is what we call path object, right? And for every path object we now have a feedback token?

pathfinding_service.rst
```diff
@@ -161,7 +166,8 @@ Example
             "path": ["0xalice", "0xeve", "0xdave", "0xbob"]
         },
         ...
-        ]
+        ],
```
Dominik1999: can we include fees in that example?

pathfinding_service.rst
```diff
@@ -161,7 +166,8 @@ Example
...
+        "feedback_token": "aaabbbcccdddeeefff"
     }
```
Dominik1999: the feedback token counts now for the whole list of paths, right? How do we differentiate which path worked out well?

pathfinding_service.rst
palango: Nice catch. So I thought as well that this is talking about the REST endpoints, but this is about interfaces in general. But I guess we can remove this as it doesn't give any valuable information.

pathfinding_service.rst
palango: It's described ion the table below. I'll make that clearer.

pathfinding_service.rst
palango: No, one feedback token is valid for all paths from a certain request.

pathfinding_service.rst
palango: Sure, will fix.

pathfinding_service.rst
palango: All routes get stored in the PFS with the token, when the client answers we lookup the token for the request and see if it's still valid and check against the provided route.

PR #234, opened 2019-06-18
Replace Merkle trees with lists

pirapira:
following the recent change to the contracts
https://github.com/raiden-network/raiden-contracts/pull/1062

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -231,7 +231,7 @@ Non cancellable, Non expirable.
 Invariants
 ^^^^^^^^^^
 
-- The :term:`balance proof` must contain the hash of pending lock list with the corresponding lock removed (and only this lock).
+- The :term:`balance proof` must contain the hash of the new list of pending locks, from which the unlocked lock has been removed.
 - This message is only sent after the corresponding partner has sent a :ref:`Reveal Secret message <reveal-secret-message>`.
 - The :term:`nonce` is increased by ``1`` with respect to the previous :term:`balance proof`
 - The :term:`locked amount` must decrease and the :term:`transferred amount` must increase by the amount held in the unlocked lock.
@@ -325,7 +325,7 @@ For the simplest Alice - Bob example:
 - Alice sends the ``RevealSecret`` to Bob and at this point she must assume the transfer is complete.
 - Bob receives the secret and at this point has effectively secured the transfer of ``n`` tokens to his side.
 - Bob sends a ``RevealSecret`` message back to Alice to inform her that the secret is known and acts as a request for off-chain synchronization.
-- Finally Alice sends an ``Unlock`` message to Bob. This acts also as a synchronization message informing Bob that the lock will be removed from the pending lock list and that the transferred_amount and locksroot values are updated.
+- Finally Alice sends an ``Unlock`` message to Bob. This acts also as a synchronization message informing Bob that the lock will be removed from the list of pending locks and that the transferred_amount and locksroot values are updated.
 
 **Mediated Transfer - Best Case Scenario**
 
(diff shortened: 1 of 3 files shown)


LefterisJP:
Why do we have the contract sources duplicated in the spec?

pirapira:
@LefterisJP for generating API docs.

LefterisJP:
So for this page? https://raiden-network-specification.readthedocs.io/en/latest/smart_contracts.html

I know it's a big change and not something to focus on now but perhaps it would make sense to have the `raiden-contracts` repo generate their own documentation and simply link to it at the contracts section of raiden spec if it's like this. Duplicating the contracts code all the time is just error prone.

pirapira:
@LefterisJP I think that's a reasonalbe solution but it's for the team to decide: https://github.com/raiden-network/spec/issues/235

messaging.rst
```diff
@@ -238,7 +231,7 @@ Non cancellable, Non expirable.
 Invariants
 ^^^^^^^^^^
 
-- The :term:`balance proof` merkle tree must have the corresponding lock removed (and only this lock).
+- The :term:`balance proof` must contain the hash of pending lock list with the corresponding lock removed (and only this lock).
```
czepluch: "of pending lock list" is this correct? To me it's difficult to decide if we talk about one lock in a list or a list of locks. And I am not sure what the "corresponding lock" refers to. Maybe it can be rephrased to be a bit more clear?

messaging.rst
```diff
@@ -332,7 +325,7 @@ For the simplest Alice - Bob example:
 - Alice sends the ``RevealSecret`` to Bob and at this point she must assume the transfer is complete.
 - Bob receives the secret and at this point has effectively secured the transfer of ``n`` tokens to his side.
 - Bob sends a ``RevealSecret`` message back to Alice to inform her that the secret is known and acts as a request for off-chain synchronization.
-- Finally Alice sends an ``Unlock`` message to Bob. This acts also as a synchronization message informing Bob that the lock will be removed from the merkle tree and that the transferred_amount and locksroot values are updated.
+- Finally Alice sends an ``Unlock`` message to Bob. This acts also as a synchronization message informing Bob that the lock will be removed from the pending lock list and that the transferred_amount and locksroot values are updated.
```
czepluch: Again here, I am uncertain if it's a list with only one lock or if a list of locks.

messaging.rst
```diff
@@ -352,6 +345,6 @@ In case a Raiden node goes offline or does not send the final balance proof to i
 
 **Limit to number of simultaneously pending transfers**
 
-The number of simultaneously pending transfers per channel is limited. The client will not initiate, mediate or accept a further pending transfer if the limit is reached. This is to avoid the risk of not being able to unlock the transfers, as the gas cost for this operation grows with the size of the Merkle tree and thus the number of pending transfers.
+The number of simultaneously pending transfers per channel is limited. The client will not initiate, mediate or accept a further pending transfer if the limit is reached. This is to avoid the risk of not being able to unlock the transfers, as the gas cost for this operation grows with the size of the pending lock list and thus the number of pending transfers.
```
czepluch: Same here. Please ignore this comment if you do indeed intent to call it "pending lock list" and not "pending locks list"

messaging.rst
pirapira: * "pending lock list":  I haven't thought about this.  People say "task list" not "tasks list".  Maybe "the list of pending locks" is clearer.
* "corresponding": I agree it's terrible now.  I'll give another try.

messaging.rst
pirapira: I'll change this to "the list of pending locks".

messaging.rst
pirapira: Same here.

messaging.rst
czepluch: Thanks. That's way better :)

messaging.rst
czepluch: Thanks. There are more places with "pending lock list" please change all of them now that there is a better term for it.

pirapira: Alright.

pirapira: Done all over.  I searched for "list lock" and I don't see anything.

PR #238, opened 2019-07-09
Updates LockedTransfer message

kelsos:
- Updates LockedTransfer message and provides step by step example
@Dominik1999

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -85,26 +85,79 @@ Messages
 Locked Transfer
 -----------------
 
-Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the list of pending locks from the sending participant node.
+A LockedTransfer is a message used to reserve tokens for a new mediated transfer.
 
-Invariants
-^^^^^^^^^^
+Preconditions
+^^^^^^^^^^^^^
 
-Only valid if all the following hold:
+For a ``LockedTransfer`` to be considered valid the following conditions have to be true.
 
 - There is a channel which matches the given :term:`chain id`, :term:`token network` address, and :term:`channel identifier`.
 - The corresponding channel is in the open state.
 - The :term:`nonce` is increased by ``1`` in respect to the previous :term:`balance proof`
 - The :term:`locksroot` must change, the new value must be equal to the root of a new tree, which has all the previous locks plus the lock provided in the message.
-- The :term:`locked amount` must increase, the new value is equal to the old value plus the lock's amount.
-- The lock's amount must be smaller then the participant's :term:`capacity`.
+- Have the new lock represented in merkleroot.
+- The :term:`locked amount` must increase by exactly ``lock.amount`` otherwise, the message would be rejected by the recipient. If the ``locked_amount`` is increased by more, then funds may get locked in the channel. If the ``locked_amount`` is increased by less, then the recipient will reject the message as it may mean it received the funds with an on-chain unlock. The initiator will stipulate the fees based on the available routes and incorporate it in the lock's amount. Note that with permissive routing it is not possible to predetermine the exact `fee` amount, as the initiator does not know which nodes are available, thus an estimated value is used..
+- The lock's amount must be smaller than the current :term:`capacity`. If the amount is higher then the recipient will reject it, as it means he will be spending money it does not own. .
 - The lock expiration must be greater than the current block number.
 - The :term:`transferred amount` must not change.
 
-Fields
-^^^^^^
+Matrix message
+^^^^^^^^^^^^^^
+
+On the transport level ``LockedTransfer`` is encoded as a JSON message. The message then might hop through a different
+number of nodes until it reaches the ``target``. 
+
+Let's assume that there is a network:
+
+- [A] `0x540B51eDc5900B8012091cc7c83caf2cb243aa86`  
+- [B] `0x2A915FDA69746F515b46C520eD511401d5CCD5e2`
+- [C] `0x811957b07304d335B271feeBF46754696694b09e`
+
+Where **A** has a channel with **B** and **B** has a channel with **C**.
+
+A <---> B <---> C
+
+If **A** wants to send 10 wei of a Token(0xc778417e063141139fce010982780140aa0cd5ab) to **C** he has to first
+send a LockedTransfer with **B** (``recipient``) where **C** is the  ``target``. After receiving the message, **B** has to send a new LockedTransfer to **C**.
+
+The message that will be sent from A -> B over the matrix transport would look like this.
+
+.. code-block:: json
+
+    {
+        "type": "LockedTransfer",
+        "chain_id": 337,
+        "message_identifier": 123456,
+        "payment_identifier": 1,
+        "nonce": 1,
+        "token_network_address": "0xe82ae5475589b828D3644e1B56546F93cD27d1a4",
+        "token": "0xc778417E063141139Fce010982780140Aa0cD5Ab",
+        "channel_identifier": 1338,
+        "transferred_amount": 0,
+        "locked_amount": 10,
+        "recipient": "0x2A915FDA69746F515b46C520eD511401d5CCD5e2",
+        "locksroot": "0x607e890c54e5ba67cd483bedae3ba9da9bf2ef2fbf237b9fb39a723b2296077b",
+        "lock": {
+            "type": "Lock",
+            "amount": 10,
+            "expiration": 1,
+            "secrethash": "0x59cad5948673622c1d64e2322488bf01619f7ff45789741b15a9f782ce9290a8"
+        },
+        "target": "0x811957b07304d335B271feeBF46754696694b09e",
+        "initiator": "0x540B51eDc5900B8012091cc7c83caf2cb243aa86",
+        "fee": 0,
+        "signature": "0x33b336f151f9790f40287655bd412a043be83a03d0136ef5e002229dd04d5b4c2b505b65911251b2a2eb428403de394064bdae0cd8d4a3bb47a10b1a0d924b921c"
+    }
+
+Message Fields
+^^^^^^^^^^^^^^
 
 This should correspond to `the packed format of LockedTransfer <https://github.com/raiden-network/raiden/blob/d504ed25b85eea5738fd3d2149bd8392a2b02226/raiden/encoding/messages.py#L164>`_.
+
+Let's call this structure of message fields ``message_structure`` from now on. Also let's assume that there is 
+a function called ``pack(message)`` that takes a this ``message_structure`` or any similar structure and returns
+a byte array.
 
 +-----------------------+----------------------+------------------------------------------------------------+
 | Field Name            | Field Type           |  Description                                               |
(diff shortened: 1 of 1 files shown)


messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
 Locked Transfer
 -----------------
 
-Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the list of pending locks from the sending participant node.
+A LockedTransfer is a message used to reserve tokens for a new mediated transfer.
```
Dominik1999: - maybe "used to reserve (to lock) tokens for a mediated transfer"
- do we have a link to an explanation of the term "mediated transfer"?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 
-Invariants
-^^^^^^^^^^
+Preconditions
+^^^^^^^^^^^^^
 
-Only valid if all the following hold:
+For a ``LockedTransfer`` to be considered valid the following conditions have to be true.
```
Dominik1999: maybe simpler "there are the following conditions"?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 
 - There is a channel which matches the given :term:`chain id`, :term:`token network` address, and :term:`channel identifier`.
```
Dominik1999: Where is the channel? Maybe we can get even more technical here. In TokenNetwork.sol or in the internal representation of the graph? On the other hand maybe that is too much ...

Dominik1999: since you use the term "corresponding" in the next bullet, maybe use "there is a channel corresponding to the given 'chain-id' ..."

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 - The corresponding channel is in the open state.
 - The :term:`nonce` is increased by ``1`` in respect to the previous :term:`balance proof`
```
Dominik1999: - is it "with respect" or "in respect"? 
- is the last possible nonce always from a balance proof or can it also be from another locked transfer?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 - The :term:`locksroot` must change, the new value must be equal to the root of a new tree, which has all the previous locks plus the lock provided in the message.
```
Dominik1999: - we don't have a tree anymore, right? It is more like a hash-list or so. There is certainly a specific term for that

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
-- The :term:`locked amount` must increase, the new value is equal to the old value plus the lock's amount.
-- The lock's amount must be smaller then the participant's :term:`capacity`.
+- Have the new lock represented in merkleroot.
```
Dominik1999: this and the previous point seem to represent the same precondition?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
 Locked Transfer
 -----------------
 
-Cancellable and expirable :term:`transfer`. Sent by a node when a transfer is being initiated, this message adds a new lock to the list of pending locks from the sending participant node.
+A LockedTransfer is a message used to reserve tokens for a new mediated transfer.
 
-Invariants
-^^^^^^^^^^
+Preconditions
```
Dominik1999: Can you maybe give the conditions numbers? Like PC1 for precondition 1. Does that make sense to you?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 - The :term:`nonce` is increased by ``1`` in respect to the previous :term:`balance proof`
 - The :term:`locksroot` must change, the new value must be equal to the root of a new tree, which has all the previous locks plus the lock provided in the message.
-- The :term:`locked amount` must increase, the new value is equal to the old value plus the lock's amount.
-- The lock's amount must be smaller then the participant's :term:`capacity`.
+- Have the new lock represented in merkleroot.
+- The :term:`locked amount` must increase by exactly ``lock.amount`` otherwise, the message would be rejected by the recipient. If the ``locked_amount`` is increased by more, then funds may get locked in the channel. If the ``locked_amount`` is increased by less, then the recipient will reject the message as it may mean it received the funds with an on-chain unlock. The initiator will stipulate the fees based on the available routes and incorporate it in the lock's amount. Note that with permissive routing it is not possible to predetermine the exact `fee` amount, as the initiator does not know which nodes are available, thus an estimated value is used..
```
Dominik1999: To me only the first sentence seems to be a precondition. The rest seems to be an explanation of the precondition and our fee model. Maybe we can find another place for that?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+- The lock's amount must be smaller than the current :term:`capacity`. If the amount is higher then the recipient will reject it, as it means he will be spending money it does not own. .
```
Dominik1999: Like capacity of my own channel view or in the channel view of my channel partner. 
There seems to be one "." too much at the end 
Only the first sentence seems to be a precondition. Even if it is nice to explain those. Maybe we can find a consistent way in either explaining every precondition or no one. Does that make sense?

messaging.rst
kelsos: I am not sure, I will have to check.

messaging.rst
kelsos: I am under the impression that it was removed, but @hackaugusto @pirapira can provide more information about that.

messaging.rst
kelsos: If you think it's better. The text here was either in the original spec or copied from the python class docstrings.

messaging.rst
kelsos: If you think it makes it simpler and more clear I would say let's go for it.

messaging.rst
Dominik1999: Do you think it is better? I think it would be more consistent, to list the preconditions and either explain them all or explain the process / reasons for it in another place. If you like to do this together, you could seperate those in the bullet itself. 
- (PC5) The :term:`locked amount` must increase by exactly ``lock.amount`` otherwise, the message would be rejected by the recipient. - Rationale: _If the ``locked_amount`` is increased by more, then funds may get locked in the channel. If the ``locked_amount`` is increased by less, then the recipient will reject the message as it may mean it received the funds with an on-chain unlock._

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 - The lock expiration must be greater than the current block number.
 - The :term:`transferred amount` must not change.
 
-Fields
-^^^^^^
+Matrix message
```
Dominik1999: Maybe we can put the message specification before the preconditions? We refer in the preconditions section to the various data fields but did not introduce them. Or is there a reason why you put that in this order?

Dominik1999: Is it clear what we mean by Matrix message? Maybe we can define that as the message and clarify in the first sentence that this message is sent over our Matrix communication Layer

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+^^^^^^^^^^^^^^
+
+On the transport level ``LockedTransfer`` is encoded as a JSON message. The message then might hop through a different
```
Dominik1999: is it the transport level or the transport layer?

Dominik1999: Is it the same message that hops through the network or all the time a slightly different message? The message gets altered with every hop, right?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+- [A] `0x540B51eDc5900B8012091cc7c83caf2cb243aa86`  
+- [B] `0x2A915FDA69746F515b46C520eD511401d5CCD5e2`
+- [C] `0x811957b07304d335B271feeBF46754696694b09e`
+
+Where **A** has a channel with **B** and **B** has a channel with **C**.
+
+A <---> B <---> C
+
+If **A** wants to send 10 wei of a Token(0xc778417e063141139fce010982780140aa0cd5ab) to **C** he has to first
+send a LockedTransfer with **B** (``recipient``) where **C** is the  ``target``. After receiving the message, **B** has to send a new LockedTransfer to **C**.
```
Dominik1999: Maybe LockedTransfer Message to make that more clear?

Dominik1999: he has to first send a LockedTransfer ~~with~~ **to** **B*, right?

Dominik1999: maybe: ~~where~~ **C** specified (or declared) as ~~is~~ the  ``target``

messaging.rst
Dominik1999: some lines down you say "After receiving the message, **B** has to send a new LockedTransfer to **C**." But I see what you mean. Somehow the information passes through the nodes but not the message itself

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+
+The message that will be sent from A -> B over the matrix transport would look like this.
+
+.. code-block:: json
```
Dominik1999: I love that you have included consistent examples

messaging.rst
kelsos: I think you can have footnotes and terms in rst so we could make the description of the preconditions a simple summary, as simple as possible as long as it makes sense and then we can have a reference to an explanation. Do you think such an approach would make sense?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+        "signature": "0x33b336f151f9790f40287655bd412a043be83a03d0136ef5e002229dd04d5b4c2b505b65911251b2a2eb428403de394064bdae0cd8d4a3bb47a10b1a0d924b921c"
+    }
+
+Message Fields
+^^^^^^^^^^^^^^
 
 This should correspond to `the packed format of LockedTransfer <https://github.com/raiden-network/raiden/blob/d504ed25b85eea5738fd3d2149bd8392a2b02226/raiden/encoding/messages.py#L164>`_.
 
+Let's call this structure of message fields ``message_structure`` from now on. Also let's assume that there is 
+a function called ``pack(message)`` that takes a this ``message_structure`` or any similar structure and returns
```
Dominik1999: that takes **a** this - there seems to be a typo?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+        "fee": 0,
+        "signature": "0x33b336f151f9790f40287655bd412a043be83a03d0136ef5e002229dd04d5b4c2b505b65911251b2a2eb428403de394064bdae0cd8d4a3bb47a10b1a0d924b921c"
+    }
+
+Message Fields
+^^^^^^^^^^^^^^
 
 This should correspond to `the packed format of LockedTransfer <https://github.com/raiden-network/raiden/blob/d504ed25b85eea5738fd3d2149bd8392a2b02226/raiden/encoding/messages.py#L164>`_.
 
+Let's call this structure of message fields ``message_structure`` from now on. Also let's assume that there is
```
Dominik1999: why didn't you call that message structure in the first place? Or is it the message structure that consists of various message fields?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
+        "target": "0x811957b07304d335B271feeBF46754696694b09e",
+        "initiator": "0x540B51eDc5900B8012091cc7c83caf2cb243aa86",
+        "fee": 0,
+        "signature": "0x33b336f151f9790f40287655bd412a043be83a03d0136ef5e002229dd04d5b4c2b505b65911251b2a2eb428403de394064bdae0cd8d4a3bb47a10b1a0d924b921c"
+    }
+
+Message Fields
+^^^^^^^^^^^^^^
 
 This should correspond to `the packed format of LockedTransfer <https://github.com/raiden-network/raiden/blob/d504ed25b85eea5738fd3d2149bd8392a2b02226/raiden/encoding/messages.py#L164>`_.
```
Dominik1999: Can we maybe put this under the message fields? Maybe first explain / list the message fields and then explain that this needs to be packed?

messaging.rst
```diff
@@ -85,27 +85,80 @@ Messages
...
 
+Let's call this structure of message fields ``message_structure`` from now on. Also let's assume that there is 
+a function called ``pack(message)`` that takes a this ``message_structure`` or any similar structure and returns
+a byte array.
+
 +-----------------------+----------------------+------------------------------------------------------------+
 | Field Name            | Field Type           |  Description                                               |
 +=======================+======================+============================================================+
```
Dominik1999: "one byte", "three bytes" and "65 bytes" seem to specify the message size whereas the other indicators seem to specify the type (solidity type, right?) of the field. Maybe clarify also the former ones as their specific type?

messaging.rst
rakanalh: The below should not be matrix-specific so i agree with leaving out "matrix"

messaging.rst
pirapira: "the root of a new tree" should be "the hash of a new list of pending locks".  Sorry I haven't changed this (I was only searching for the word "Merkle").

PR #240, opened 2019-07-10
Add spec for lock expiry

rakanalh:


rakanalh:
Thanks @czepluch i will delay merging this until @Dominik1999 is done with his PR

PR #241, opened 2019-07-11
Spec for withdraw messages

rakanalh:


messaging.rst
```diff
@@ -276,6 +276,147 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+
+- The channel for which the withdraw is requested should be open.
+- The ``total_withdraw`` value should only ever increase.
+- The participant's channel balance should be larger or equal to ``withdraw_amount``,
+  which is calculated using ``new_total_withdraw - previous_total_withdraw``.
+- The new total_withdraw value should not cause an overflow.
+- The message should be sent by one of the channel participants.
+- The :term:`nonce` is increased by ``1`` with respect to the previous :term:`nonce`
+- The ``signature`` must be from the :term:`sender` of the request.
+
```
err508: Is `sender == participant` required?  I'd mention that here

messaging.rst
```diff
@@ -276,6 +276,147 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+-------------
+
+Message used by the a channel participant node to request the other participant signature on a new increased ``total_withdraw`` value.
+
+Preconditions
+^^^^^^^^^^^^^
+
+- The channel for which the withdraw is requested should be open.
+- The ``total_withdraw`` value should only ever increase.
+- The participant's channel balance should be larger or equal to ``withdraw_amount``,
```
err508: balance excluding currently locked amount?

messaging.rst
```diff
@@ -276,6 +276,147 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+  which is calculated using ``new_total_withdraw - previous_total_withdraw``.
+- The new total_withdraw value should not cause an overflow.
```
err508: or underflow

messaging.rst
rakanalh: Done

messaging.rst
rakanalh: Good catch... added `unlocked balance`

BOR4:
using mix of "should" and "must" confuses me a little. It sounds to me like each "should" is not a hard condition but an option while "must" is a condition which has to be met. If that is the case then all good.

messaging.rst
```diff
@@ -276,6 +276,148 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+
+Preconditions
+^^^^^^^^^^^^^
+
+- The channel for which the withdraw is requested should be open.
+- The ``total_withdraw`` value should only ever increase.
+- The participant's channel unlocked balance should be larger or equal to ``withdraw_amount``,
+  which is calculated using ``new_total_withdraw - previous_total_withdraw``.
+- The new total_withdraw value should not cause an underflow or overflow.
+- The message should be sent by one of the channel participants.
```
err508: As this is a require in the contracts, I'd use must here.

messaging.rst
```diff
@@ -276,6 +276,148 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+|  participant                  | address       | The address of the withdraw requesting node                    |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  total_withdraw               | uint256       | The new monotonic ``total_withdraw`` value                     |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  expiration                   | uint256       | The block number at which withdraw request is no longer        |
+|                               |               | usable on-chain.                                               |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  nonce                        | uint64        | See :ref:`balance-proof-offchain`                              |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  signature                    | bytes         | Elliptic Curve 256k1 signature                                 |
```
err508: I think instead of which curve is being used, what data is actually signed would be more informative here.

messaging.rst
```diff
@@ -276,6 +276,148 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+Withdraw Confirmation
+-------------
+
+Message used by the :ref:`withdraw-request-message` receiver to confirm the request after validating it's input.
+
+Preconditions
+^^^^^^^^^^^^^
+
+- The channel for which the withdraw is confirmed should be open.
+- The received confirmation should map to a previously sent request.
```
err508: How does the mapping work? Is the Withdraw Confirmation an envelope for Withdraw Request?

messaging.rst
```diff
@@ -276,6 +276,148 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+- The block at which withdraw expires should not have been reached.
```
err508: Would the client confirm a withdraw request with expiration n+1, where n is the current block?

messaging.rst
```diff
@@ -276,6 +276,148 @@ This should match `the Secret message in encoding/messages file <https://github.
...
+|  participant                  | address       | The address of the withdraw requesting node                    |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  total_withdraw               | uint256       | The new monotonic ``total_withdraw`` value                     |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  expiration                   | uint256       | The block number at which withdraw request is no longer        |
+|                               |               | usable on-chain.                                               |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  nonce                        | uint64        | See :ref:`balance-proof-offchain`                              |
++-------------------------------+---------------+----------------------------------------------------------------+
+|  signature                    | bytes         | Elliptic Curve 256k1 signature                                 |
```
err508: Also, more information on what is being signed would help I think

messaging.rst
rakanalh: No not really. This is implementation specific where for the "initiator" of the withdraw, a map of all requests should be kept so that the withdraw confirmation should confirm to an existing key in the map.

messaging.rst
rakanalh: Yes. However, the requestor will reject this and prevent itself from sending a transaction because it knows that the transaction will fail due to the expiration block being reached.

messaging.rst
rakanalh: Done

messaging.rst
rakanalh: Done

messaging.rst
rakanalh: done

PR #242, opened 2019-07-11
Update LockedTransfer Spec

Dominik1999:


messaging.rst
```diff
@@ -85,71 +85,305 @@ Messages
...
+
+- (PC1) :term:`chain id`, :term:`token network` address, and :term:`channel identifier` refer to a existing channel
+- (PC2) The corresponding channel is in the open state
+- (PC3) :term:`nonce` is increased by ``1`` with respect to the previous message in that direction, e.g.
+:term:`balance proof` (IS THIS CORRECT?)
+- (PC4) :term:`locksroot` must be equal to the hash of a new list of all currently pending locks, always the latest one
+appended at last position
+- (PC5) :term:`locked amount` must increase by exactly ``lock.amount`` [#PC5]_
+- (PC6) The lock's amount (WHAT IS THAT?) must be smaller than the current :term:`capacity` [#PC6]_
+- (PC7) The lock expiration must be greater than the current block number (IS THAT A DATA FIELD? WHY NOT LINKED?)
```
kelsos: So after reading this comment and checking, I realized that my data are kind of wrong since I put `1` in expiration.  The value of `expiration` is the number of the block where the `Lock` expires. So it makes sense that if the current block, for example, is `10395313` the expiration cannot be at the same block but has to be in a future block. Does that clarify the question?

messaging.rst
```diff
@@ -85,71 +85,305 @@ Messages
...
+For a ``LockedTransfer`` to be considered valid there are the following conditions. The message will be rejected otherwise:
+
+- (PC1) :term:`chain id`, :term:`token network` address, and :term:`channel identifier` refer to a existing channel
+- (PC2) The corresponding channel is in the open state
+- (PC3) :term:`nonce` is increased by ``1`` with respect to the previous message in that direction, e.g.
+:term:`balance proof` (IS THIS CORRECT?)
+- (PC4) :term:`locksroot` must be equal to the hash of a new list of all currently pending locks, always the latest one
+appended at last position
+- (PC5) :term:`locked amount` must increase by exactly ``lock.amount`` [#PC5]_
+- (PC6) The lock's amount (WHAT IS THAT?) must be smaller than the current :term:`capacity` [#PC6]_
```
kelsos: Let's say I want to send 10 tokens. As part of the process, I have to create a lock `Lock` of 10 tokens and add it to the pending locks. Then I calculate the locksroot for the pending locks and use the locksroot in the process of creating the signature. So if my total capacity is 8 I am practically unable to lock 10 tokens to send.

messaging.rst
```diff
@@ -85,71 +85,305 @@ Messages
...
+- (PC7) The lock expiration must be greater than the current block number (IS THAT A DATA FIELD? WHY NOT LINKED?)
+- (PC8) :term:`transferred amount` must not change (COMPARED TO WHAT?)
```
kelsos: Assume that you have already transferred 10 tokens. This is your transferred amount. When sending a new `LockedTransfer` the transferred amount does not change. Only the `locked_amount` changes. For the `transferred_amount` you have to receive an `Unlock` message after the target successfully receives the secret and the `Unlock` successfully reaches the initiator.

PR #246, opened 2019-07-24
Fix the build of docs for read the docs

LefterisJP:
So the spec docs have not been built for almost 5 months ... we should really add a simple travis CI or something that builds the docs and fails a PR if it can't build the docs.


karlb:
The build on master fails locally for me, too. But the docs on https://raiden-network-specification.readthedocs.io/en/latest/ are up to date, so apparently they are building on readthedocs. Where do you see that they "not been built for almost 5 months"?

LefterisJP:
here: https://readthedocs.org/projects/raiden-network-specification/builds/?page=1 and here:

https://readthedocs.org/projects/raiden-network-specification/builds/?page=1

2019-07-26-115423_887x811_scrot
2019-07-26-115431_930x426_scrot

karlb:
The builds time out. So maybe the latest html doc got published despite the timeout because it was created early during the build?

This still leaves the question why we don't get the same syntax errors on readthedocs that we get locally. I'll try to compare the environments.

karlb:
OK, I can explain why the build succeeds (apart from the timeout) on readthedocs:

* Current sphinx supports the `code-block`. We should pin sphinx to a specific version so that it at least behaves the same for everyone building locally.
* Our makefile uses `-W` to turn warnings into errors, which is not done by readthedocs. That's the reason it does not fail on short underlines, missing terms and refs with missing backticks.

messaging.rst
```diff
@@ -60,7 +60,7 @@ Invariants
 - Expiration must be larger than the current block number and smaller than the channel’s settlement period.
 
 Hash
-^^^^
```
karlb: Why did you change this?

LefterisJP:
> The code-block removal is unnecessary (updating sphinx should solve those errors), but AFAIK the result is the same both and the changed version is more compact, so that seems preferable.

In their docs they seemed to suggest to use code-block *only* for specific languages. So I would follow their docs.

messaging.rst
LefterisJP: My thing complained locally. Even though it does not seem to be smaller.

PR #256, opened 2019-09-03
Reflect Message ID differentiation for closing and updating

pirapira:
This follows the implementation change in
https://github.com/raiden-network/raiden-contracts/pull/1150

Since Balance Proof Update message is used both for closeChnanel()
and updateNonClosingBalanceProof() call, the signature must be
differentiated using different message IDs.


pirapira:
Good idea.  I'll add a sentence or two.

pirapira:
@palango I added more about message IDs.  Please have another look.

Issue #253, opened 2019-09-03
Make sure Raiden Network Smart Contracts Specification has the correct information, data fields and tables


pirapira:
This issue keeps track of updating the spec for the current version of Raiden Network Smart Contracts.
This is a part of #252.

There have been several ABI breaking changes recently.  I first make a list of those breaking changes and see if each change is reflected in the spec.

pirapira:
## Unreleased

## [0.31.0](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.31.0) - 2019-08-20

- [ ] [#1163](https://github.com/raiden-network/raiden-contracts/pull/1163) MonitoringService.monitor() no longer works for service providers that are not registered in ServiceRegistry
- [ ] [#1177](https://github.com/raiden-network/raiden-contracts/pull/1177) When ServiceRegistry is deprecated, deposits can be immediately be withdrawn.

## [0.30.0](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.30.0) - 2019-08-15

- [ ] [#1150](https://github.com/raiden-network/raiden-contracts/pull/1150) Use different message ID for chnanelClose() and updateNonClosingBalanceProof()
- [ ] [#1148](https://github.com/raiden-network/raiden-contracts/pull/1148) Add TokenNetworkRegistry's address to MonitoringService's constructor arguments
- [ ] [#1160](https://github.com/raiden-network/raiden-contracts/pull/1160) Revert the upgrade web3.py and other dependencies
- [ ] [#1151](https://github.com/raiden-network/raiden-contracts/pull/1151) OneToN doesn't work for service providers not registered in ServiceRegistry

## [0.29.1](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.29.1) - 2019-08-13

- [ ] [#1153](https://github.com/raiden-network/raiden-contracts/pull/1153) Upgrade web3.py and other dependencies

## [0.29.0](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.29.0) - 2019-07-26

- [ ] [#1143](https://github.com/raiden-network/raiden-contracts/pull/1143) Add ServiceRegistry.hasValidRegistration() function that returns whether a given address has a valid registration.

## [0.28.0](https://github.com/raiden-network/raiden-contracts/releases/tag/v0.28.0) - 2019-07-26

[...]

PR #265, opened 2019-09-05
Add description of the deprecation switch

pirapira:
I wanted to add a new event about the deprecation switch, but
the deprecation was never explained in the spec.

This commit is a part of updating the spec according to the implementation
https://github.com/raiden-network/spec/issues/253

### DIFF
--- a/smart_contracts.rst
+++ b/smart_contracts.rst
@@ -712,6 +712,23 @@ Unlocks all pending transfers by providing all pending transfers data. The hash 
     ``unlock`` must be called after ``settleChannel`` because it needs the ``locksroot`` from the latest :term:`balance proof` in order to guarantee that all locks have either been unlocked or have expired.
 
 
+.. _turn_on_deprecation_switch:
+
+**Turning on the deprecation switch**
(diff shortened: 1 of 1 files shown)


smart_contracts.rst
```diff
@@ -712,6 +712,23 @@ Unlocks all pending transfers by providing all pending transfers data. The hash
     ``unlock`` must be called after ``settleChannel`` because it needs the ``locksroot`` from the latest :term:`balance proof` in order to guarantee that all locks have either been unlocked or have expired.
 
 
+.. _turn_on_deprecation_switch:
+
+**Turning on the deprecation switch**
+
+Allows the deprecation executor to turn on the deprecation switch. After this no channels accept new deposits, and no new channel can be opened.
```
palango: *turn on the deprecation switch* sounds a bit strange. What about *to deprecate the contract*?

pirapira: That sounds better.

pirapira: Changed the sentence so.

PR #266, opened 2019-09-05
Add expiration_block to setTotalWithdraw() args

pirapira:
This commit adds the new argument `expiration_block` to
`setTotalWithdraw()` function.

The change follows the implementation
https://github.com/raiden-network/raiden-contracts/pull/1103

This commit is a part of updating the spec
https://github.com/raiden-network/spec/issues/253


smart_contracts.rst
```diff
@@ -477,6 +478,7 @@ Allows a channel participant to withdraw tokens from a channel without closing i
 - ``total_withdraw``: Total amount of tokens that are marked as withdrawn from the channel during the channel lifecycle.
 - ``participant_signature``: Elliptic Curve 256k1 signature of the channel ``participant`` on the :term:`withdraw proof` data.
 - ``partner_signature``: Elliptic Curve 256k1 signature of the channel ``partner`` on the :term:`withdraw proof` data.
+- ``expiration_block``: the block number when the withdraw message becomes no longer valid.
```
palango: Is this:
- The last block that's valid
- The first block that's invalid

pirapira: It's going to be "the first block number when the withdraw message is no longer valid".

PR #275, opened 2019-09-06
Update messages specification for Alderaan

jomuel:
The message format tables are still not up-to-date. There will be a follow-up for that.

Some changes to the messaging specification:
- add an Overview section (as is planned for all chapters of the spec)
- add some missing message and data structure descriptions
- structure the text better
- elaborate more on the structure of the balance proof
- move the transfer specification into an own chapter

### DIFF
--- a/mediated_transfer.rst
+++ b/mediated_transfer.rst
@@ -4,18 +4,22 @@ Mediated transfers
 Overview
 ========
 
-The protocol supports mediated transfers. Nodes may use them to send payments.
-A :term:`Mediated transfer` may be cancelled and can expire until the initiator reveals the secret.
+Nodes can use a :term:`mediated transfer` to send payments to another node without opening a
+direct channel to it.
 
-A mediated transfer is done in two stages, possibly on a series of channels:
+A mediated transfer is done in two stages, usually through multiple channels:
 
-- Reserve token :term:`capacity` for a given payment, using a :ref:`locked transfer message <locked-transfer-message>`.
-- Use the reserved token amount to complete payments, using the :ref:`unlock message <unlock-message>`
+- **Allocation:** Reserve token :term:`capacity` for a given payment, using a
+  :ref:`locked transfer message <locked-transfer-message>`.
+- **Finalization:** Use the reserved token amount to complete payments, using the
+  :ref:`unlock message <unlock-message>`
+
+A mediated transfer may be cancelled and can expire until the initiator reveals the secret.
 
 Mediated Transfers
 ==================
 
-A :term:`Mediated Transfer` is a hash-time-locked transfer. Currently raiden supports only one type of lock. The lock has an amount that is being transferred, a :term:`secrethash` used to verify the secret that unlocks it, and a :term:`lock expiration` to determine its validity.
+A :term:`mediated transfer` is a hash-time-locked transfer. Currently raiden supports only one type of lock. The lock has an amount that is being transferred, a :term:`secrethash` used to verify the secret that unlocks it, and a :term:`lock expiration` to determine its validity.
 
 Mediated transfers have an :term:`initiator` and a :term:`target` and a number of mediators in between. The number of mediators can also be zero as these transfers can also be sent to a direct partner. Assuming ``N`` number of mediators, a mediated transfer will require ``10N + 16`` messages to complete. These are:
 
@@ -42,17 +46,17 @@ For the simplest Alice - Bob example:
 - Bob sends a ``RevealSecret`` message back to Alice to inform her that the secret is known and acts as a request for off-chain synchronization.
 - Finally Alice sends an ``Unlock`` message to Bob. This acts also as a synchronization message informing Bob that the lock will be removed from the list of pending locks and that the transferred_amount and locksroot values are updated.
 
-Mediated Transfer - Best Case Scenario
---------------------------------------
+Mediated Transfer - Happy Path Scenario
+---------------------------------------
 
-In the best case scenario, all Raiden nodes are online and send the final balance proofs off-chain.
+In the happy path scenario, all Raiden nodes are online and send the final balance proofs off-chain.
 
 .. image:: diagrams/RaidenClient_mediated_transfer_good.png
     :alt: Mediated Transfer Good Behaviour
     :width: 900px
 
-Mediated Transfer - Worst Case Scenario
----------------------------------------
+Mediated Transfer - Unhappy Path Scenario
+-----------------------------------------
 
 In case a Raiden node goes offline or does not send the final balance proof to its payee, then the payee can register the ``secret`` on-chain, in the ``SecretRegistry`` smart contract before the ``secret`` expires. This can be used to ``unlock`` the lock on-chain after the channel is settled.
 
--- a/mediated_transfer.rst
+++ b/mediated_transfer.rst
@@ -30,21 +30,26 @@ Mediated transfers have an :term:`initiator` and a :term:`target` and a number o
 - ``2N + 3`` processed (one for everything above)
 - ``5N + 8`` delivered
 
-For the simplest Alice - Bob example:
+For a simple example with one mediator:
 
-- Alice wants to transfer ``n`` tokens to Bob.
+- Alice wants to transfer ``n`` tokens to Charlie, using Bob as a mediator.
 - Alice creates a new transfer with:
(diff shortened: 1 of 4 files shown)


mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
+Mediated transfers
+##################
+
+Overview
+========
+
+The protocol supports mediated transfers. Nodes may use them to send payments.
```
Dominik1999: very good idea to outsource this chapter. Here, why you don't have `term: Mediated Transfer` like the lines below?

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+A :term:`Mediated transfer` may be cancelled and can expire until the initiator reveals the secret.
```
Dominik1999: can you move this sentence below? I would expect first the explanation about what a mediated transfer is and then how it can fail

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+
+A mediated transfer is done in two stages, possibly on a series of channels:
```
Dominik1999: Isn't every mediated transfer done over a series of channels? Can you find a better term than series of channels? Isn't Multi-hop the term used when it comes to Raiden / Lightning?

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+
+- Reserve token :term:`capacity` for a given payment, using a :ref:`locked transfer message <locked-transfer-message>`.
```
Dominik1999: Can we put 
- **Allocation:** reserves token ...
- **Finalization:** use the reserved token amount either to unlock or ...

mediated_transfer.rst
Dominik1999: ... ok I see in the example below we can also do it with hops ... For a better understanding, can we re-write the example to a mediated transfer with a hop?

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+- Use the reserved token amount to complete payments, using the :ref:`unlock message <unlock-message>`
+
+Mediated Transfers
+==================
+
+A :term:`Mediated Transfer` is a hash-time-locked transfer. Currently raiden supports only one type of lock. The lock has an amount that is being transferred, a :term:`secrethash` used to verify the secret that unlocks it, and a :term:`lock expiration` to determine its validity.
+
+Mediated transfers have an :term:`initiator` and a :term:`target` and a number of mediators in between. The number of mediators can also be zero as these transfers can also be sent to a direct partner. Assuming ``N`` number of mediators, a mediated transfer will require ``10N + 16`` messages to complete. These are:
```
Dominik1999: nice description

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+
+- ``N + 1`` :term:`locked transfer` or :term:`refund transfer` messages
+- ``1`` :term:`secret request`
+- ``N + 2`` :term:`reveal secret`
+- ``N + 1`` :term:`unlock`
+- ``2N + 3`` processed (one for everything above)
+- ``5N + 8`` delivered
+
+For the simplest Alice - Bob example:
```
Dominik1999: can we create an example, maybe even consistent with the example in messaging.rst, that has a mediator?

mediated_transfer.rst
```diff
@@ -0,0 +1,71 @@
...
+Mediated Transfer - Best Case Scenario
+--------------------------------------
+
+In the best case scenario, all Raiden nodes are online and send the final balance proofs off-chain.
+
+.. image:: diagrams/RaidenClient_mediated_transfer_good.png
+    :alt: Mediated Transfer Good Behaviour
+    :width: 900px
+
+Mediated Transfer - Worst Case Scenario
```
Dominik1999: when we have `best case` and `worst case` scenarios, isn't it somehow implicit that there are many cases in between? If so, should we list all those cases? If not, and I don't see more cases atm, can you rename those?

messaging.rst
```diff
@@ -4,7 +4,36 @@ Raiden Messages Specification
 Overview
 ========
 
-This is the specification document for the messages used in the Raiden protocol.
+This is the specification of the messages used in the Raiden protocol.
+
+There are data structures which reappear in different messages:
+The :ref:`offchain balance proof <balance-proof-offchain>`
```
Dominik1999: can you have those 2 in a numbered list?

messaging.rst
```diff
@@ -4,7 +4,36 @@ Raiden Messages Specification
 Overview
 ========
 
-This is the specification document for the messages used in the Raiden protocol.
+This is the specification of the messages used in the Raiden protocol.
```
Dominik1999: Can yu maybe add something like. There are x different messages in raiden. we use y for the coordination of a mediated transfer and z for the channel withdraw

mediated_transfer.rst
jomuel: As I understand it, multi-hop means two or more mediators, so not all mediated transfers are multi-hop.

messaging.rst
jomuel: @karlb suggested a different classification of the messages that will be in the followup pr.

messaging.rst
jomuel: I did a bullet-point list to be consistent with the other overview sections.

mediated_transfer.rst
jomuel: Good point, maybe we call the two cases "happy path" and "unhappy path"?

mediated_transfer.rst
Dominik1999: ok I see. Then we should not use the word multi-hop. 

A mediated transfer is a transfer where we need to have at least one hop, right? If there on no hop (e.g. in a direct transfer) we do not need a mediator and therefore, the transfer is not mediated. 
So I would need in the specification, 
- the explanation and definition of what a mediated transfer is 
- a consistent example that reflects this definition

Would you agree to that?

messaging.rst
Dominik1999: ok, sure

messaging.rst
Dominik1999: ah ok, then ignore my comment.

mediated_transfer.rst
jomuel: My intuition would also have been that a transfer with no mediators is not considered a mediated transfer. However, the spec in its old state stresses that there can be "mediated transfers" with no mediators, and it even gave an example with no mediator (the Alice-Bob example we changed). The point is probably that the mediated transfer protocol with ``LockedTransfer``, ``SecretRequest`` etc messages still works with no mediators (and is currently the way we do direct transfers?)

mediated_transfer.rst
```diff
@@ -0,0 +1,80 @@
...
+  :ref:`unlock message <unlock-message>`
+
+A mediated transfer may be cancelled and can expire until the initiator reveals the secret.
+
+Mediated Transfers
+==================
+
+A :term:`mediated transfer` is a hash-time-locked transfer. Currently raiden supports only one type of lock. The lock has an amount that is being transferred, a :term:`secrethash` used to verify the secret that unlocks it, and a :term:`lock expiration` to determine its validity.
+
+Mediated transfers have an :term:`initiator` and a :term:`target` and a number of mediators in between. The number of mediators can also be zero as these transfers can also be sent to a direct partner. Assuming ``N`` number of mediators, a mediated transfer will require ``10N + 16`` messages to complete. These are:
```
Dominik1999: - Can we somehow say that this "The number of mediators can also be zero as these transfers can also be sent to a direct partner." is an edge case and put it to the end of the description
- is it true that in this edge case, this ``10N + 16`` would mean that ther are 16 messages needed to complete a mediated transfers between two nodes without an mediator?

mediated_transfer.rst
Dominik1999: yes that is true. But it is still not a mediated transfer imho. We just use the same protocol for direct transfers as we use it for mediated transfers. But using the same protocol doesn't make it the same think. We shouldn't stress that point tbh, because it makes it only more complicated for the reader without adding value. Thanks for creating the other example!

mediated_transfer.rst
jomuel: I singled it out as a note. As far as I can tell there are indeed 16 messages in that case.

PR #276, opened 2019-09-11
WIP: Add spec for mediation fee calculation

palango:
WIP


mediation_fees.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. math::
+
+    b(1+p) = b + bq + qb(1+p)  \\
+    (1+p) = 1 + q + q(1+p)  \\
+    q + q(1+p) = p \\
+    q(2+p) = p \\
+    q = \frac{p}{2+p}
+
+Forward calculation (as in the client)
```
Dominik1999: for every fee combined now, right? so maybe make this with the same heading-level as the last chapter

mediation_fees.rst
```diff
@@ -0,0 +1,89 @@
...
+=====================================================================
+
+User usually think about mediation fees as a atomic actions, so it's easier to
+let them define a *per-hop* proportional mediation fee (called :math:`p`).
+However, this setting then needs to be converted into a *per-channel*
+proportional fee (called :math:`q`).
+
+.. math::
+
+    b(1+p) = b + bq + qb(1+p)  \\
```
Dominik1999: where does this `b + bq + qb(1+p)` come from?

Dominik1999: for readability: can you change `qb(1+p)` to `bq(1+p)` ?

Dominik1999: equation is correct, I also get q = p/(2+p)

PR #279, opened 2019-09-13
Messaging specs update for Alderaan, part 2

jomuel:
This is the second and last part of the message/transfer spec update.

- Introduce a message classification
- Update of packing format for hashes and signatures
- New example
- Put mediated transfer chapter before messaging chapter

**Edit:** There will be a follow-up PR to add the missing format of `LockedTransfer.metadata`.


terminology.rst
```diff
@@ -206,3 +212,17 @@ Raiden Terminology
 
    expiration
         Specific block after which the lock in the :term:`locked transfer` expires
+
+   offchain message
+        A message that is only sent between Raiden nodes, with no information that could be
+        forwarded to a contract - for example, a ``Processed`` message.
+
+   onchain message
```
karlb: I'm still afraid people might be misled by that name. What do you think of "on-chain capable message"? Not great either, but at least somewhat clearer. I'd like someone else to chime in on this before merging it.

terminology.rst
```diff
@@ -206,3 +212,17 @@ Raiden Terminology
 
    expiration
         Specific block after which the lock in the :term:`locked transfer` expires
+
+   offchain message
```
karlb: The usual spelling seems to be "off-chain" and "on-chain".

terminology.rst
rakanalh: IMO, there's no such thing as an on-chain message. I would suggest removing the whole off-chain / on-chain message.

All messages are off-chain. Transactions are executed onchain using certain signatures / data exchanged via messages.

>         A message that is sent between Raiden nodes but could as well be forwarded to a contract.

This is incorrect.

> Notice that a message does not have to be published on a contract to be considered an
        onchain message

Doesn't make much sense really. 

> It only has to be of a type that allows sending it to a contract in total, e. g. ``WithdrawRequest``.

hmm what type is that? 

As i said, we don't need this definition as trying to fit the word `onchain` and `message` is invalid.

karlb: There are certain sets of messages that are signed in different ways for good reasons. It will be easier to talk about the differences when we give them some name or description, even if it's something long like "Messages that contain a signature that is necessary to create a specific on-chain transaction".

At least for the signatures, there are three different groups of messages. If we can't say anything about these groups of messages in general without getting inaccurate, we should at least write up a proper description on how the signatures work and why they are constructed in the way they are.

terminology.rst
```diff
@@ -206,3 +212,17 @@ Raiden Terminology
...
+
+   onchain message
+        A message that is sent between Raiden nodes but could as well be forwarded to a contract.
+        Notice that a message does not have to be published on a contract to be considered an
+        onchain message - It only has to be of a type that allows sending it to a contract in
+        total, e. g. ``WithdrawRequest``.
+
+   envelope message
+        A message that partly consists of information which may be forwarded to a contract later,
+        such as a :term:`balance proof`. An example is the ``LockedTransfer`` message.
```
karlb: We're using `pack_balance_proof` in `EnvelopeMessage._data_to_sign`, so if I'm not missing anything, it is *always* a balance proof.

jomuel:
@rakanalh 
I wasn't happy with the term `on-chain message` myself, and changed it to `publishable message` now. That is still awkward, but not misleading (`on-chain message` could suggest that it always goes on-chain, not that its information can potentially be published on-chain). 

What I am trying to document here is that there are three subclasses of Raiden messages with different packing/signing/hashing conventions. I think this information belongs in the messaging spec. 
(Though maybe not in the glossary)

messaging.rst
```diff
@@ -159,202 +159,107 @@ Locked Transfer message
...
+| fee                                  | uint256 |  32         |
++--------------------------------------+---------+-------------+
+| metadata_hash                        | bytes32 |  32         |
++--------------------------------------+---------+-------------+
+
+The computation of the ``metadata_hash`` as well as the exact format of the ``metadata`` itself
+are implementation specific.
+
+This will be used to generate the the data field called ``additional_hash``, which is a required
+part of the process to create the message signature. It is computed as the ``keccak256``-hash
```
karlb: I'm not sure where to put it, but it would be nice to have an explanation somewhere why we have the `additional_hash` in the first place. The reason is that it allows as to notice if any field has been tampered with (because the `additional_hash` would be different and thus the overall signature) while still having a signature that will be accepted by the contract (because it uses the additional hash only for the signature check and nothing else).

Other approaches would require two signatures, which would make this process less efficient.

messaging.rst
```diff
@@ -504,19 +504,27 @@ Preconditions
 - The message sender address must be the same as ``participant``.
 - The ``signature`` must be from the :term:`sender` of the request.
 
-Fields
-^^^^^^
+Fields and signature
+^^^^^^^^^^^^^^^^^^^^
+
+``WithdrawRequest`` is an :term:`onchain message`. This is the format in which it is packed to
+compute the ``signature``, and in which it could be forwarded to a contract.
```
karlb: The `WithdrawRequest` itself is never sent to the blockchain, but its signature will be used (along with the signature of the `WithdrawConfirmation`) to withdraw tokens by calling `setTotalWithdraw` on the token network contract.

messaging.rst
```diff
@@ -610,31 +609,34 @@ Preconditions
 Fields
 ^^^^^^
 
+``WithdrawExpired`` is an :term:`onchain message`. This is the format in which it is packed to
+compute the ``signature``, and in which it could be forwarded to a contract.
```
karlb: None of the fields of `WithdrawExpired` ever go to the contract. Which raises the question why the signature is constructed in the way that the contracts expect. Is this just for consistency with the other two withdraw messages, or is there a different reason @rakanalh?

messaging.rst
```diff
@@ -159,202 +159,107 @@ Locked Transfer message
...
++--------------------------------------+---------+-------------+
+| amount                               | uint256 |  32         |
++--------------------------------------+---------+-------------+
+| fee                                  | uint256 |  32         |
++--------------------------------------+---------+-------------+
+| metadata_hash                        | bytes32 |  32         |
++--------------------------------------+---------+-------------+
+
+The computation of the ``metadata_hash`` as well as the exact format of the ``metadata`` itself
+are implementation specific.
```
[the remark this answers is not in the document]
jomuel: Good question at least for multihop transfers, since the metadata contains the routing information and mediator nodes will not route themselves. So the metadata format is part of the protocol and has to be documented in all it's awkwardness?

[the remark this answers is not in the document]
karlb: I've not been involved in any discussions about this format, but yes, I think so.

messaging.rst
jomuel: I thought that explanation was already in `terminology.rst`, will check.

terminology.rst
```diff
@@ -196,7 +202,7 @@ Raiden Terminology
 
    additional hash
    additional_hash
```
karlb: `additional_hash` is duplicated on the above two lines.

messaging.rst
```diff
@@ -22,31 +23,56 @@ The further messages within the transfer are based on it:
...
+  balance proof in turn contains an :term:`additional hash`, which is a hash over the rest of
+  the message. Each envelope message has a defined packed data format to compute the additional
+  hash. The format always starts with the 1-byte command id. Envelope messages are:
+  ``LockedTransfer``, ``RefundTransfer``, ``Unlock`` and ``LockExpired``.
+
+- The second group is messages that will never result in on-chain transactions, as they contain
+  no information that could be forwarded to a contract. There are four types of such messages,
+  which we will call **internal messages**: ``SecretRequest``, ``RevealSecret``, ``Processed``
+  and ``Delivered``. Internal messages have a packed data format in which they are signed.
+  As with envelope messages, the format always starts with the message type's 1-byte command id,
```
karlb: > As with envelope messages

Actually this is somewhat different from envelope messages. For envelope messages, the data for their `additional_hash` starts with the cmd_id. While for internal messages, the data for the signature starts with the cmd_id.

messaging.rst
```diff
@@ -22,31 +23,56 @@ The further messages within the transfer are based on it:
...
+  but unlike them, the command id is followed by a padding of three zero bytes.
+
+- In addition there are the withdraw-related message types: ``WithdrawRequest``,
+  ``WithdrawConfirmation``, and ``WithdrawExpired``. They have a packed formats starting with the
+  :term:`token network address`, the :term:`chain id` and a message type constant, which is an
+  unsigned 256-bit integer. The signature obtained from this format are used in some contract
```
karlb: We can be more specific here without using many more words:

> The signatures from ``WithdrawRequest`` and ``WithdrawConfirmation`` are used when withdrawing tokens in the ``TokenNetwork`` contract.

messaging.rst
```diff
@@ -22,31 +23,56 @@ The further messages within the transfer are based on it:
...
+
+- **Envelope messages**, which contain a balance proof which can be sent to a contract. The
+  balance proof in turn contains an :term:`additional hash`, which is a hash over the rest of
+  the message. Each envelope message has a defined packed data format to compute the additional
+  hash. The format always starts with the 1-byte command id. Envelope messages are:
+  ``LockedTransfer``, ``RefundTransfer``, ``Unlock`` and ``LockExpired``.
+
+- The second group is messages that will never result in on-chain transactions, as they contain
+  no information that could be forwarded to a contract. There are four types of such messages,
+  which we will call **internal messages**: ``SecretRequest``, ``RevealSecret``, ``Processed``
```
karlb: You could object to the term "internal messages" by saying that they are transferred between different nodes and are thus not internal. But I don't have a better name and it is unambiguous enough when used in this context, so I don't object.

PR #280, opened 2019-09-13
Define a goal for the specs

jomuel:
(Up for discussion)

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -2,7 +2,18 @@ Raiden Network Specification
 =================================
 
 
-This is a tentative specification for the `Raiden Network <http://raiden.network/>`_. It's under development and will be further refined in subsequent iterations.
+This is the specification of the `Raiden Network <http://raiden.network/>`_. Its goal is to provide
+an exact description of the Raiden protocol.
+
+For the Raiden clients (Raiden nodes), we will thus describe the format of messages it sends to other
+raiden nodes, and the structure of a transfer composed of such messages. For implementation specific
+information on the
+`reference implementation of the Raiden client <https://github.com/raiden-network/raiden/>`_, like
+its API, refer to the `docs <https://raiden-network.readthedocs.io/en/stable/index.html>`_ instead.
+
+Components that the client relies on, such as the pathfinding service, monitoring service and
+Raiden smart contracts, are specified here such that client implementations can be developed
+against them, i. e. including a specified API.
 
 .. toctree::
    :maxdepth: 2


index.rst
```diff
@@ -2,7 +2,19 @@ Raiden Network Specification
 =================================
 
 
-This is a tentative specification for the `Raiden Network <http://raiden.network/>`_. It's under development and will be further refined in subsequent iterations.
+This is the specification of the `Raiden Network <http://raiden.network/>`_.
+
+Its goal is to provide an exact description of the Raiden protocol. So the structure of
+a mediated transfer, message formats and such are to be found here, while implementation
```
rakanalh: What does `and such` mean in this case? do we have anything else that describes the protocol other than messages?

rakanalh:
```suggestion
a mediated transfer, message formats and such are to be found here. Implementation-specific
```

index.rst
```diff
@@ -2,7 +2,19 @@ Raiden Network Specification
...
+specific parts of the Raiden client, which is the protocol's reference implementation, are
+documented in the `docs <https://raiden-network.readthedocs.io/en/stable/index.html>`
+instead.
+
+Smart contracts and services, however, are specified in greater detail here since every possible
```
rakanalh:
```suggestion
Smart contracts and services are also specified in greater detail here since every possible
```

index.rst
```diff
@@ -2,7 +2,19 @@ Raiden Network Specification
...
+implementation of Raiden needs to be developed against them. This makes it necessary, for example,
+to specify the API endpoints for the services, whereas the API of the Raiden client is not documented
+here but in the Raiden docs.
+
```
rakanalh: `This makes it necessary, for example,
to specify the API endpoints for the services, whereas the API of the Raiden client is not documented
here but in the Raiden docs.` 

Sounds like a justification, please remove.

index.rst
```diff
@@ -2,7 +2,17 @@ Raiden Network Specification
 =================================
 
 
-This is a tentative specification for the `Raiden Network <http://raiden.network/>`_. It's under development and will be further refined in subsequent iterations.
+This is the specification of the `Raiden Network <http://raiden.network/>`_. Its goal is to provide
+an exact description of the Raiden protocol.
+
+For the Raiden client (Raiden node), we will thus describe the format of messages it sends to other
```
pirapira: Let's talk about plural "Raiden clients" instead of "the Raiden client".  The point is then clearer.

jomuel: Good idea.

PR #281, opened 2019-09-17
Update imbalance fee usage

karlb:
This reflects the changes done in
https://github.com/raiden-network/raiden/pull/4901

### DIFF
--- a/mediation_fees.rst
+++ b/mediation_fees.rst
@@ -53,7 +53,7 @@ Fee calculation
 
 There are two fundamental formula to relate :math:`a`, :math:`b` and :math:`c`.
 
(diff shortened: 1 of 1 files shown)


mediation_fees.rst
```diff
@@ -53,7 +53,7 @@ Fee calculation
 
 There are two fundamental formula to relate :math:`a`, :math:`b` and :math:`c`.
 
-1. :math:`a = c + aq + f + i(a)`
+1. :math:`a = c + aq + f + i(-a)`
```
palango: A small comment here to explain why `i(-a)` is used would be nice, but we can also do this later

PR #302, opened 2019-09-25
Add summary to transport specification and update/add some information.

err508:


transport.rst
```diff
@@ -91,19 +102,12 @@ Peer discovery room
...
-
-Direct Communication Rooms
-''''''''''''''''''''''''''
-In Matrix, users can send direct e2e encrypted messages to each other through private/invite-only rooms.
-
-Blockchain Event Rooms
-''''''''''''''''''''''
-Each operator registered with the Raiden :ref:`ServiceRegistry` could provide a room, where relevant events from Raiden Token Networks are published. E.g. signed, so that false info could be challenged.
-
+Raiden Nodes publish BalanceUpdates to the Pathfinding Service room. The Pathfinding Service can
```
Dominik1999: I think the nodes here send two different messages PFSCapacityUpdate and PFSFeeUpdates or so. Can you re-check with Paul or Karl?

transport.rst
```diff
@@ -71,17 +84,15 @@ Matrix allows to check for the presence of a user. Clients should listen for cha
 Sending transfer messages to other nodes
 ----------------------------------------
 
-Direct Message, which is modeled as a room with 2 participants.
-Channel rooms have the name in the format ``raiden_<network_name_or_id>_<peerA>_<peerB>``, where ``peerA`` and ``peerB`` are sorted in lexical-order. As the users may roam to other homeservers, participants should keep listening for user-join events in the presence room, if it's a user of interest (with which it shares a room), with valid signed ``displayName`` and not yet in the room we share with it, invite it to the room.
 
+Private channel rooms random room_ids in the format ``random_room_id:homeserver``, where the
```
jomuel:
```suggestion
Private channel rooms have random room ids in the format ``random_room_id:homeserver``, where the
```

We could consider the random ids as an implementation detail not relevant for the spec, since (afaik) only the convenience room ids mentioned below are used?

transport.rst
```diff
@@ -1,16 +1,26 @@
 Transport
 #########
 
+Summary
+============
+
+Raiden is a network agnostic protocol. Protocol messages can in general be transferred over
```
jomuel: It is true that the network itself is transport-agnostic, but we should make clear here that Matrix is the only transport option of the reference implementation, so to be able to join the existing/evolving Raiden Network a client needs to speak Matrix.

transport.rst
err508: I added it. correct @karlb ?

transport.rst
err508: No it's the other way around, we only use random room_ids and give the rooms aliases afterwards

transport.rst
err508: Added.

transport.rst
karlb: Yes, looks good, now.

PR #320, opened 2019-10-11
Explain calculation of default IP function

karlb:
The formula is already used in the Raiden client. This supplies the
reasoning why the values are chosen as they are.

### DIFF
--- a/mediation_fees.rst
+++ b/mediation_fees.rst
@@ -224,7 +224,7 @@ Requirements
 ------------
 
 In order to make it easier to enable imbalance fees, the Raiden client
-includes a default IP function that can be configured by a single
+includes a default imbalance penalty (IP) function that can be configured by a single
 parameter (``--proportional-imbalance-fee`` on the Raiden CLI).
 
 The function is chosen to have the following properties:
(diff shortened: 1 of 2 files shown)


mediation_fees.rst
```diff
@@ -215,3 +215,73 @@ Now forward and backward calculation should let us recalculate :math:`b` or :mat
...
+
+Derivation of :math:`a` and :math:`b`
+-------------------------------------
+
+Starting with the function formula and its derivative
+
+.. math::
+   \begin{align}
+   f(x) &= a|x-o|^b \\
+   f'(x) &= ab(x-o)|o-x|^{b-2}
```
Dominik1999: wolfram alpha say the same. I did not believe that this might be the first derivative. Nice

mediation_fees.rst
```diff
@@ -215,3 +215,73 @@ Now forward and backward calculation should let us recalculate :math:`b` or :mat
...
+   \end{align}
+
+as well as
+
+.. math::
+   f(0) := c \quad\text{and}\quad f'(0) := -s
```
Dominik1999: where does the s come from? Ah the slope. Can you put that in brackets after that statement?

mediation_fees.rst
```diff
@@ -215,3 +215,73 @@ Now forward and backward calculation should let us recalculate :math:`b` or :mat
...
+   \end{align}
+
+and :math:`b`
+
+.. math::
+   \begin{align}
+   f'(0) &= -s \\
+   ab(-o)o^{b-2} &= -s \\
+   abo^{b-1} &= s \\
+   \frac{c}{o^b}bo^{b-1} = \frac{cb}{o} &= s \\
```
Dominik1999: Can you add in brackets (including a)

PR #337, opened 2021-02-02
Add capabilites docs to transport page

palango:
As taken from https://github.com/raiden-network/raiden/issues/6503#issuecomment-704310156

Part of https://github.com/raiden-network/team/issues/882

### DIFF
--- a/transport.rst
+++ b/transport.rst
@@ -156,4 +156,51 @@ offline. The Monitoring Service will submit their balance proof on their behalf.
 Pathfinding Service Updates Room
 ''''''''''''''''''''''''''''''''
 Raiden Nodes publish :term:`PFSCapacityUpdate` and :term:`PFSFeeUpdate` to the Pathfinding Service room. The Pathfinding Service can
- compute efficient routes throughout the network and provide these routes to requesting nodes.
+compute efficient routes throughout the network and provide these routes to requesting nodes.
+
+Capabilities
+------------
+
+Raiden clients need a way to signal their capabilities to other nodes. This is done by encoding the capabilities in the ``avatar_url`` field of the user profile.
+
+Serialization for use in ``avatar_url``
+'''''''''''''''''''''''''''''''''''''''
+
+The following template is used to encode the capabilities in the avatar URL field:
+::
+
+    mxc://raiden.network/cap?{capabilities_url_encoded}
+
+Here ``{capabilities_url_encoded}`` is the url query parameter encoding of the capabilities.
+
+Rules for url encoding:
+
+* boolean values are encoded as truthy values, e.g. ``"0"`` and ``"1"``
+* other values are encoded as strings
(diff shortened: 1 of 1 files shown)


transport.rst
```diff
@@ -123,4 +123,51 @@ offline. The Monitoring Service will submit their balance proof on their behalf.
...
+The following template is used to encode the capabilities in the avatar URL field:
+::
+
+    mxc://raiden.network/cap?{capabilities_url_encoded}
+
+Here ``{capabilities_url_encoded}`` is the url query parameter encoding of the capabilities.
+
+Rules for url encoding:
+
+* bool values are encoded as ``"0"`` for ``False`` and ``"1"`` for ``True``
```
andrevmatos, line 145: I'd prefer the spec doesn't state explicitly encoding for boolean values. The LC, for example, does support decoding `[Ff]alse` and `[Tt]rue` values as boolean, and decimal numeric strings as numbers. Instead, I'd push for these (or even less) explicit decoding rules, and then for the specific code which depends/handles specific capabilities to interpret the truthness or some other interpretation of these values.
Again, in the LC case, the `1` and `0` strings are decoded as numbers, and their truthness implicitly converts them to boolean, but I'd have that not required by spec.

andrevmatos: Maybe true/false strings and numeric strings would be sensible to be in the spec. If not desired, then plain strings only, and then have each capability code explicitly interpreting these values as better as they can (since capabilities are a best-effort and non-binding in any way).

palango: > I'd prefer the spec doesn't state explicitly encoding for boolean values.

But the above stuff is what you agreed on with PC team!?

andrevmatos: Kinda.. it's a little bit implementation specific. They did it like that, but in the call, I told I'd prefer a more generic approach, which would be compatible between both implementations, so we did it. But now, to state in the spec, I'd prefer the more generic one, since both clients should both accept it and output something which is compatible anyway.

palango: What about:

> boolean values are encoded as ``"0"`` for ``False`` and ``"1"`` for ``True``, but existing clients accept `[Ff]alse` and `[Tt]rue` as well.

andrevmatos: I'm not sure PC accepts that (the LC does). Maybe word it something like _existing implementations encode boolean values as  falsy or truthy values, like "0" and "1", therefore it's expected that new compatible implementations are capable of decoding these parameters as falsy and truthy values as well, as numbers or booleans_. I think the idea is to state that this is the current behavior, but Capabilities, by definition, are non-binding and a best-effort protocol, therefore I proposed the more generic decoding behavior only to be required by spec.

palango: I adapted the text and added a note to the decoding section.

andrevmatos (approved):
lgtm

PR #338, opened 2021-02-02
Update message sending mechanisms

palango:
As described in https://hackmd.io/zAx2LQjOQteO-yX4m3dG-A

Part of https://github.com/raiden-network/team/issues/882


transport.rst
```diff
@@ -101,12 +101,43 @@ Matrix allows to check for the presence of a user. Clients should listen for cha
...
+Matrix ``toDevice`` messages
+''''''''''''''''''''''''''''
+Matrix also supports ``toDevice`` messages. These are not stored permanently as
+part of a shared communication history and are delivered exactly once to each
+client device.
+
+As Raiden does not rely on the messaging history, this is a valid alternative
+and usually shows lower latency and reduces load on the Matrix federation nodes.
+
+Nodes that support ``toDevice`` messages signal that with the ``toDevice``
```
andrevmatos, line 130: Maybe state that clients compatible with this capability are expected to have a `deviceId="RAIDEN"`, and clients sending this type of messages may specify `RAIDEN` as target deviceId.

transport.rst
```diff
@@ -101,12 +101,43 @@ Matrix allows to check for the presence of a user. Clients should listen for cha
...
+capability.
+
+WebRTC messaging
+''''''''''''''''
+To further optimize the communication, exchanging messages in a peer-to-peer
+manner is possible with WebRTC. In that case matrix is only used to initiate the
+WebRTC connection (either via private rooms, or ``toDevice`` messages).
+
+Nodes that support WebRTC messages signal that with the ``WebRTC`` capability.
```
andrevmatos, line 139:
```suggestion
Nodes that support WebRTC messages signal that with the ``webRTC`` capability.
```

transport.rst
```diff
@@ -101,12 +101,43 @@ Matrix allows to check for the presence of a user. Clients should listen for cha
...
+
```
andrevmatos, line 140: The WebRTC signaling protocol is a little specific. Should we include it here, or somewhere else?

palango: I tried to keep it high level, but we can add it if you like.

andrevmatos: I don't think it needs to be done now, but just to keep it in mind.

andrevmatos (approved):
lgtm

PR #340, opened 2021-04-22
Add withdraw diagram 

karlb:
We didn't add one when we added withdraw to the protocol. I added support for mermaid diagrams to do this easily.

See https://raiden-network-specification--340.org.readthedocs.build/en/340/smart_contracts.html#withdraw for the preview.


palango (approved):
Nice to have those diagrams generated inline 🥳

PR #347, opened 2021-10-15
Add changes to metadata

ezdac:
This mainly adds the documentation for the newly employed additions to the Metadata to the messages.
It will give some insight into why this changes are necessary and explain the secret encryption as well 
as document some encryption and encoding schemes related to the mentioned changes.

### DIFF
--- a/messaging.rst
+++ b/messaging.rst
@@ -119,13 +119,13 @@ revealing the secret via the additionally required :ref:`RevealSecret <reveal-se
 The secret should be encrypted with the public key of the LT's target, using the ECIES encryption scheme (look e.g. `here <https://ecies.org/>`__ for reference implementations).
 The encrypted data has to be encoded as an object containing secret and additional metadata needed for validation::
 
-        secret = encrypt_ecies(target_publickey, { "secret": <bytes32>, "amount": <int32>, "payment_identifier": <int32> })
+        secret = encrypt_ecies(target_publickey, { "secret": <bytes32>, "amount": <uint256>, "payment_identifier": <uint64> })
 
-Note the encoding of the encrypted object's field is not part of this specification.
+Note the encoding of the encrypted object's fields is not part of this specification.
 
 When the target receives a transfer with an encrypted secret, it may try to decrypt it::
 
-        { "secret": <bytes32>, "amount": <int32>, "payment_identifier": <int32> } = encrypt_ecies(target_privatekey, secret)
+        { "secret": <bytes32>, "amount": <uint256>, "payment_identifier": <uint64> } = encrypt_ecies(target_privatekey, secret)
 
 
 The target should only accept the decrypted secret, if the decrypted objects fields comply to the following properties:


messaging.rst
```diff
@@ -104,38 +104,113 @@ Metadata
...
+
+The target should only accept the decrypted secret, if the decrypted objects fields comply to the following properties:
+
+- ``secret`` must hash to the LT's ``lock.secrethash``
+- ``amount`` must be less than or equal to the received LT's ``lock.amount``
+- ``payment_identifier`` must be equal to LT's ``payment_identifier``
+
+If the validation passes, the target uses "secret" as this transfer's secret and skips sending a :ref:`SecretRequest <secret-request-message>` to the initator.
+Therefore the target can immediately start the unlock phase by sending a corresponding :ref:`RevealSecret <reveal-secret-message>` to the last mediator in the transfer's route.
+
```
andrevmatos, line 139:
```suggestion
Otherwise, the usual ``SecretRequest`` protocol should be performed as a fallback; even in unrecoverable cases like not enough `amount` (meaning mediators took more than initiator expected/were willing to pay), sending the received-by-target payment info back to initiator in ``SecretRequest``'s payload is useful to inform initiator if and why this transfer failed, and allow them to notify user.

```

messaging.rst
```diff
@@ -104,38 +104,113 @@ Metadata
...
+The Metadata should be determined by the initiator. All mediators of a transfer should pass it to the next hop of the transfer as-is and
+should not modify it. Complying to this behaviour is advertised as the :ref:`capability <transport-capabilities>` `immutableMetada=1`.
```
andrevmatos, line 141:
```suggestion
should not modify it even though unknown fields are present. Complying to this behaviour is advertised as the :ref:`capability <transport-capabilities>` `immutableMetada=1`.
```

messaging.rst
```diff
@@ -104,38 +104,113 @@ Metadata
...
 
 .. _route-metadata:
 
-This datastructure describes a single route, the noce receiving the message
-will be the first entry in the list, the second node is the node which the
-transfer should be forwarded to.
+This datastructure describes a single route and the metadata for all nodes included in that route.
 
-Each entry is EIP55-checksum addresses with ``0x``-prefix as usual. The last of
+Each entry in the ``route`` field list is EIP55-checksum addresses with ``0x``-prefix as usual. The last of
```
andrevmatos, line 161: We had removed address checksum format as a hard-requirement on the protocol, and instead coded the clients to accept any format plus as-is signature verification, which would allow any supported format to interoperate. Should we adapt the spec here?

messaging.rst
```diff
@@ -201,13 +276,13 @@ The data will be packed as follows to compute the :term:`additional hash`:
 | metadata_hash                        | bytes32 |  32         |
 +--------------------------------------+---------+-------------+
 
-The ``metadata_hash`` is defined using `RLP <https://github.com/ethereum/wiki/wiki/RLP>`__.
+
+The ``metadata_hash`` is defined using `JCS <https://datatracker.ietf.org/doc/html/rfc8785>`__.
 It is given as::
```
andrevmatos, line 281:
```suggestion
The ``metadata_hash`` is defined using `JCS <https://datatracker.ietf.org/doc/html/rfc8785>`__, right after JSON parsing (i.e. before any decoding/transformation is performed). It is given as::
```

messaging.rst
ezdac: Yes, this makes sense

messaging.rst
```diff
@@ -104,38 +104,113 @@ Metadata
...
+to be encrypted in a specific way, so that only the target of the LT can decrypt it.
+If the encrypted secret is included in the ``Metadata``, the transfer can't be cancelled by the initiator by not 
+revealing the secret via the additionally required :ref:`RevealSecret <reveal-secret-message>` message.
+
+..
+
+The secret should be encrypted with the public key of the LT's target, using the ECIES encryption scheme (look e.g. `here <https://ecies.org/>`__ for reference implementations).
+The encrypted data has to be encoded as an object containing secret and additional metadata needed for validation::
+
+        secret = encrypt_ecies(target_publickey, { "secret": <bytes32>, "amount": <int32>, "payment_identifier": <int32> })
```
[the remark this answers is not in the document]
andrevmatos: Yes, I agree, we should be consistent on the notations here, but at least the number sizes could/should be fixed, just to not lead to a wrong implementation; other than that, should be good to go

messaging.rst
```diff
@@ -104,38 +104,113 @@ Metadata
...
+
+Note the encoding of the encrypted object's field is not part of this specification.
+
+When the target receives a transfer with an encrypted secret, it may try to decrypt it::
+
+        { "secret": <bytes32>, "amount": <int32>, "payment_identifier": <int32> } = encrypt_ecies(target_privatekey, secret)
```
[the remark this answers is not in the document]
andrevmatos: :thumbsup:

PR #348, opened 2021-10-15
Update TokenNetwork contract

ezdac:
- added the `TokenNetwork.openChannelWithDeposit` to the contracts
- renamed a mention of the deprecation executor for the TokenNetwork contract

### DIFF
--- a/smart_contracts.rst
+++ b/smart_contracts.rst
@@ -480,6 +480,34 @@ Deposit more tokens into a channel. This will only increase the deposit of one o
 
     A participant or a delegate ``MUST`` be able to deposit more tokens into a channel, regardless of his partner's availability.
 
+
+.. _open-deposit-channel:
+
+Open channel with funding
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


smart_contracts.rst
```diff
@@ -480,6 +480,34 @@ Deposit more tokens into a channel. This will only increase the deposit of one o
...
+    )
+        public
+
+
+- ``participant1``: Ethereum address of a channel participant
+- ``participant2``: Ethereum address of the other channel participant
+- ``settle_timeout``: Number of blocks that need to be mined between a call to closeChannel and settleChannel
+- ``participant1_total_deposit``: The total amount of tokens that ``participant1`` will have as deposit
+
+As the :ref:``depositChannel <deposit-channel>`` function, a successful call to this function will emit a `ChannelNewDeposit` event in the same way.
```
andrevmatos, line 509: Suggestion:
```suggestion
As the :ref:``depositChannel <deposit-channel>`` function, a successful call to this function will emit a `ChannelNewDeposit` event in the same way and same block (so clients must ensure they're able to react to deposit events inclusive on the block when `ChannelOpen` was detected).
```

PR #349, opened 2021-10-18
Update transport specification

ezdac:
Since our transport layer and it's integration with the services had quite some changes during the last months,
the spec has to be fundamentally updated.
Some changes included:

- more responsibility indirected from home-server to PFS (presence-hook, no explicit presence in clients anymore )
- (discovery-)room removal
- switch to toDevice as "default" for P2P (node to node)
- toDevice broadcast implementation for services

### DIFF
--- a/transport.rst
+++ b/transport.rst
@@ -173,6 +173,34 @@ nodes is handled over a webRTC data-channel.
 Nodes that support WebRTC messages signal this functionality with the ``webRTC`` capability.
 The requirement is optional, but lack thereof will reduce transfer speeds significantly.
 
+To establish a WebRTC data-channel with another peer, the clients are expected to:
+
+#. Get peer's presence/address-metadata information, either from a PFS's specific endpoint or from a passing-through ``LockedTransfer``
+#. Verify that the peer has ``webRTC`` capability
+#. Signaling messages are sent to peer's ``userId`` using ``toDevice`` matrix messages, with ``type`` as ``m.room.message`` (even though it's not in a Matrix room) and ``content`` as ``{ "msgtype": "m.notice", "body": <payload> }``, where ``payload`` is the JSON-encoded string of an object in the format ``{ "type": <signal_type>, "call_id": <dataChannel.label>, ...<rest of payload> }``
+#. In parallel, start a call of its own, and also listen for calls/offers from peers of interest:
+
+   #. On the caller side:
+
+      #. create a ``RTCPeerConnection`` and a ``dataChannel`` on it, with whatever label is desired, to uniquely identify this channel upon related messages (e.g. ``<0xCaller_address>|<0xCallee_address>|<timestamp>``)
+      #. start listening for ``ICECandidates`` on this connection, and send them to peer, with ``type="candidates"`` and a ``candidates`` payload member containing an array with the gathered candidates
+      #. create an ``offer``, set it as ``local description`` on connecting, and send it to the peer with ``type="offer"`` and a ``sdp`` payload member containing the offer string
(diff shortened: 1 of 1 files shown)


transport.rst
```diff
@@ -81,82 +81,98 @@ Example:
...
+
+WebRTC messaging
+''''''''''''''''
+To further optimize the communication, exchanging messages in a peer-to-peer
+manner is possible with WebRTC. In that case matrix is only used to initiate the
+WebRTC connection via P2P ``toDevice`` messages, and successive communication between 
+nodes is handled over a webRTC data-channel.
+
+Nodes that support WebRTC messages signal this functionality with the ``webRTC`` capability.
+The requirement is optional, but lack thereof will reduce transfer speeds significantly.
```
andrevmatos, line 175: Maybe we should add our custom WebRTC signaling protocol here? Maybe something like
```suggestion

To establish a WebRTC data-channel with another peer, the clients are expected to:
1. Get peer's presence/address-metadata information, either from PFS's specific endpoint or from a passing-through `LockedTransfer`, and has `webRTC` capability
2. Signaling messages are sent to peer's `userId` using `toDevice` matrix messages, with `type` as `m.room.message` (even though it's not in a room) and `content` as `{ "msgtype": "m.notice", "body": <payload> }`, where `payload` is the JSON-encoded string of an object in the format `{ "type": <signal_type>, "call_id": <dataChannel.label>, ...<rest of payload> }`
3. In parallel, start a call of its own, and also listen for calls/offers from peers of interest:
  - On the caller side:
    1. create a `RTCPeerConnection` and a `dataChannel` on it, with whatever label is desired, to uniquely identify this channel upon related messages; an example: `0xCaller_address|0xCallee_address|timestamp`;
    2. start listening for `ICECandidates` on this connection, and send them to peer, with `type="candidates"` and a `candidates` payload member containing an array with the gathered candidates;
    3. create an `offer`, set it as `local description` on connectin, and send it to peer, with `type="offer"` and a `sdp` payload member containing the offer string;
    4. wait for an `answer` message from peer, and upon receiving it, set it as `remote description`; `RTCDataChannel` should then become `open`
    5. a timeout may be put, to retry if neither this call nor callee's side managed to get a channel opened
  - On the callee side:
    1. listen for `offer` messages
    2. when receiving one from a peer of interest, create an `RTCPeerConnection` and set `offer` as `remote description`;
    3. start listening for `ICECandidates` on this connection, and send it to peer, the same as on caller's side
    3. create `answer`, set it as `local description` and send it to peer, with `type="answer"` and `sdp` payload member containing the answer string
    4. wait for `dataChannel` to be emitted and to become `open`
    5. listening for offers on callee is permanent, and any new offer coming through, if succeed, may disconnect previous callee or caller channels;
4. Both caller and callee's codepaths can race; the winner of this race (first channel to become `open`) for each peer's (by address) will disconnect the other direction, and this now-open `RTCDataChannel` will be kept and used for this partner's messaging
5. Upon channel error or close, peers may send a `type="hangup"` message, without additional payload members, and then possibly retry the loop above;
6. Clients may retry whenever they want, if the peer is online; any new open RTC channel for each peer disconnects the previous one; usually, it's ok to retry this just a couple of times and give up, as partner seems to be offline or not responding, and assume they'll call when they come back online; additionally, they may trigger the loop again upon certain events, as new raiden `ChannelOpen` is detected or a message needs to be sent;
```

andrevmatos (approved):
Just suggested adding the RTC protocol; it may require some formatting though; on the rest, LGTM, thanks for this!

transport.rst
ezdac: Yes, perfect - thank you! I will format and add this as is (content wise) since I don't know too much about the details of our rtc-signalling

